PR #1, opened 2018-05-09
Initialize project

djydewang:
@adangel @jsotuyod 
I initialized the project with [hoe](https://github.com/seattlerb/hoe). Besides, I also added the following:

- command-line interface
- options class and its tests
- a rake task used to perform all tests

All in all, it's enough to start our work IMO.
BTW, maybe we should to add CI to the project.


lib/pmdtester/parsers/options.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    def parse(argv)
+      Slop.parse argv do |o|
+        o.string '-r', '--LocalGitRepo', 'path to the local PMD repository'
+        o.string '-b', '--baseBranch', 'name of the base branch in local PMD repository'
+        o.string '-p', '--patchBranch', 'name of the patch branch in local PMD repository'
+        o.string '-bc', '--baseConfig', 'path to the base PMD configuration file'
+        o.string '-pc', '--patchConfig', 'path to the patch PMD configuration file'
+        o.string '-c', '--config', 'path to the base and patch PMD configuration file'
+        o.string '-l', '--listOfProjects', 'path to the file which contains the list of standard projects'
```
jsotuyod: I'd change the long option names to something more standard in other command line tools: all lowercase, split with dash when needed. ie: --local-git-repo or --patch-branch

README.txt
```diff
@@ -0,0 +1,32 @@
+= pmdtester
```
jsotuyod: I'd rename this file to `README.md`, as it should be a Markdown file rather than a simple text file

History.txt
```diff
@@ -0,0 +1,6 @@
+=== 1.0.0 / 2018-05-09
```
jsotuyod: I'd change this to `CHANGELOG.md` in order for:

1. Make it clear we want markdown here, and not just plain text
2. be consistent in the naming of these files in all uppercase

lib/pmdtester/parsers/options.rb
djydewang: Done.

README.txt
djydewang: Done.

History.txt
djydewang: I tried to do this, but [hoe](http://yukimotopress.github.io/gem-tasks) didn't allow History.txt to disappear. And hoe can help create many useful rake tasks, so I think we can make a compromise here.

lib/pmdtester/parsers/options.rb
```diff
@@ -0,0 +1,55 @@
...
+      @config = options[:c]
+      @project_list = options[:l]
+      @mode = options[:m]
+    end
+
+    private
+
+    def parse(argv)
+      Slop.parse argv do |o|
+        o.string '-r', '--local-git-repo', 'path to the local PMD repository'
```
adangel: We'll need a more detailed documentation about the options later on. I assume, that you use "-r", if you don't want to clone the PMD repo but rather reuse an existing local clone and thus avoid unnecessary internet traffic.

lib/pmdtester/parsers/options.rb
```diff
@@ -0,0 +1,55 @@
...
+        o.string '-b', '--base-branch', 'name of the base branch in local PMD repository'
+        o.string '-p', '--patch-branch', 'name of the patch branch in local PMD repository'
+        o.string '-bc', '--base-config', 'path to the base PMD configuration file'
+        o.string '-pc', '--patch-config', 'path to the patch PMD configuration file'
+        o.string '-c', '--config', 'path to the base and patch PMD configuration file'
+        o.string '-l', '--list-of-projects', 'path to the file which contains the list of standard projects'
+        o.string '-m', '--mode', "the mode of the tool: 'local', 'online' or 'single'\n" +
+            "\tsingle: Set this option to 'single' if your patch branch contains changes for any option that can't work on master/base branch\n" +
```
adangel: I'm afraid, I don't understand this. Maybe we can skip this mode "single", until we need it? So, we'll only implement the features, that we need, but nothing more...

test/test_options.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def test_long_option
+    opts = PmdTester::Options.new(
+        %w[--local-git-repo /path/to/repo --base-branch pmd_releases/6.2.0 --patch-branch master --config config/all-java.xml])
+    assert_equal("/path/to/repo", opts.local_git_repo)
+    assert_equal("pmd_releases/6.2.0", opts.base_branch)
+    assert_equal("master", opts.patch_branch)
+    assert_equal("config/all-java.xml", opts.config)
+  end
+end
```
adangel: I'd add one test case with the typical minimum required options. I think, in order to call the regression tester, you need also the location of the project list (`-l`). This test case can serve as an example of how to call the tool.

lib/pmdtester/parsers/options.rb
```diff
@@ -0,0 +1,55 @@
+require 'slop'
+
+module PmdTester
+  class Options
+
+    VERSION = '1.0.0'
```
adangel: Is it possible to add a version suffix, to indicate that this version is work in progress - like "`VERSION = '1.0.0-SNAPSHOT'`. Using 1.0.0 might otherwise be misinterpreted as "finished" and "feature complete"...

Manifest.txt
```diff
@@ -0,0 +1,7 @@
+History.txt
+Manifest.txt
+README.md
+Rakefile
+bin/pmdtester
+lib/pmdtester/parsers/options.rb
+test/test_options.rb
```
adangel: Please add the file `LICENSE` there, too.

README.md
```diff
@@ -0,0 +1,31 @@
+= pmdtester
```
adangel: Since this is now a markdown file (that's the preferred format we use at PMD), the syntax is now a bit different:

The headlines start with `#` and `##` (for the 2nd level) instead of the equal signs.
You can see at https://github.com/djydewang/pmd-regression-tester-1 how github interprets this.

Rakefile
```diff
@@ -0,0 +1,15 @@
...
+
+require 'hoe'
+require './lib/pmdtester/parsers/options.rb'
+
+Hoe.spec 'pmdtester' do
+  self.version = PmdTester::Options::VERSION
+
+  self.author  = 'Binguo Bao'
+  self.email   = 'djydewang@gmail.com'
+
```
adangel: I think, with the option "history_file" we can make the changelog also using markdown. I don't have a strong preference for the name of the changelog (PMD uses release_notes.md somewhere in the docs, eclipse plugin uses ReleaseNotes.md).

I'm not sure about the correct syntax, maybe:

```
  self.history_file = 'History.md'
```

Actually, [reading the source of hoe yields](https://github.com/seattlerb/hoe/blob/master/lib/hoe.rb#L650): It will use "History.txt" or "History.md", whichever file exists. If none exists, then "History.txt" is used by default. So, if we stay with "History.md", then you just need to rename the file, if we choose a different filename entirely, then we need to configure hoe.

lib/pmdtester/parsers/options.rb
djydewang: Single mode is for new rules or new features, and we need to test the performance of new rules or new features on standard projects. The old version doesn't need to be involved in this case.

Rakefile
djydewang: > Actually, reading the source of hoe yields:

I didn't find any documents on how to set up "history file", so it seems that reading the source code is also a good choice:)

lib/pmdtester/parsers/options.rb
djydewang: > I assume, that you use "-r", if you don't want to clone the PMD repo but rather reuse an existing local clone and thus avoid unnecessary internet traffic.

Anyways, even if you're going to clone the PMD repo, the address needs to be provided.
Developers are sure to have PMD repo locally, and when we integrate to CI, the clone task should be done with other scripts rather than this tool IMO:)

test/test_options.rb
djydewang: Good catch. I missed the location of the project list.

> I'd add one test case with the typical minimum required options.

The above two test cases are the ones you expect after adding `-l` and `--list-of-project` to them,

config/all-java.xml
```diff
@@ -0,0 +1,18 @@
...
+    <description>Every Java Rule in PMD</description>
+
+    <rule ref="category/java/bestpractices.xml" />
+    <rule ref="category/java/codestyle.xml" />
+    <rule ref="category/java/design.xml" />
+    <rule ref="category/java/documentation.xml" />
+    <rule ref="category/java/errorprone.xml" />
+    <rule ref="category/java/multithreading.xml" />
+    <rule ref="category/java/performance.xml" />
+<!--     <rule ref="category/java/security.xml" /> -->
```
jsotuyod: we now have rules under security! :D https://pmd.github.io/pmd-6.3.0/pmd_rules_java_security.html

lib/pmdtester/parsers/options.rb
```diff
@@ -0,0 +1,55 @@
+require 'slop'
+
+module PmdTester
+  class Options
+
+    VERSION = '1.0.0-SNAPSHOP'
```
jsotuyod: SNAPSHOP?

lib/pmdtester/parsers/options.rb
jsotuyod: I agree with Andreas here... we should start with as little flags as we actually need, and add more as we go and see a use case arise. I'm even doubtful of the local flag (when I work from home I want to check my changes performance against a baseline... why use a local run when I have a result freely available? I can only think of some kind of "offline mode" as a valid usage for it, and I'd still consider it)

lib/pmdtester/parsers/options.rb
djydewang: I'm not sure about it. Please see [comment](https://github.com/pmd/pmd-regression-tester/pull/1#discussion_r187400561).

lib/pmdtester/parsers/options.rb
djydewang: When we implement local mode (my prototype is to implement this mode), other modes are easy to implement. The application scenarios of these three modes are predictable. When we integrate tools into CI, we certainly use online mode; when developers want to compare specific two branches/tags of PMD("result freely available" you mentioned is not present in this case), local mode is a good choice.

lib/pmdtester/parsers/options.rb
adangel: You're right - we always need to provide the repo in some ways.

It seems like Slop also supports marking options as required via `required: true`. Maybe you can mark the mandatory commandline options that way in one of the next PRs?

PR #2, opened 2018-05-14
Add projects parser & design format of projectlist

djydewang:
I will finish the TODO in projects_parser.rb after uploading the schema file to somewhere.

### DIFF
--- a/config/project-list.xml
+++ b/config/project-list.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0"?>
+
+<projectlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+xsi:noNamespaceSchemaLocation="projectlist_1_0_0.xsd">
+  <description>Standard Projects</description>
+
+  <project>
+    <name>checkstyle</name>
+    <type>git</type>
+    <connection>https://github.com/checkstyle</connection>
+    <webview-url>https://github.com/checkstyle</webview-url>
+    <tag>checkstyle-8.10</tag>
+    <exclude-pattern>**/src/test/resources-noncompilable/**/*</exclude-pattern>
+    <exclude-pattern>**/src/test/resources/**/*</exclude-pattern>
+  </project>
+
+  <project>
+    <name>spring-framework</name>
+    <type>git</type>
+    <connection>https://github.com/spring-projects/spring-framework</connection>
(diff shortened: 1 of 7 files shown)


lib/pmdtester/parsers/projects_parser.rb
```diff
@@ -0,0 +1,24 @@
...
+module PmdTester
+
+  class ProjectsParser
+    def initialize(list_file)
+      @list_file = list_file
+    end
+
+    def parse
+      list_xml_doc = Document.new(File.new(@list_file))
+      #TODO validate the XML file against an XSD schema
```
adangel: Uploading the schema file should not be needed. Usually you can refer the the xsd file the same way you refer to the xml file.
I didn't find documentation on REXML, but if you search for "ruby validate xml against xsd" you'll find some alternative libraries, e.g. https://stackoverflow.com/questions/1396759/xsd-schema-validation-in-ruby suggests to use https://rubygems.org/gems/nokogiri/ . nokogiri seems to be still under active development, while rexml is not (https://rubygems.org/gems/rubysl-rexml).

We can upload the schema file later on, as it makes it easier for external xml editors. But it is not need while working on this project - this project will always be the primary and definitive source of the schema.
That's similar to the other schema we have in PMD - while they are available from the web, the primary source is the source code (e.g. https://github.com/pmd/pmd/blob/master/pmd-core/src/main/resources/ruleset_2_0_0.xsd)

config/projectlist_1_0_0.xsd
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" ?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
```
adangel: Once the schema has matured, it makes sense to define an own namespace for it. This will make it easier, if we want to upgrade it (e.g. version 2.0.0). But that's not needed right now.

config/project-list.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0"?>
+
+<projectlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+xsi:noNamespaceSchemaLocation="projectlist_1_0_0.xsd">
```
jsotuyod: sweet! I'd love to have an XML schema for this file for ease of use

PR #3, opened 2018-05-16
Add pmd report builder

djydewang:
The PR is WIP.


lib/pmdtester/cmd.rb
```diff
@@ -0,0 +1,19 @@
...
+  class Cmd
+    def self.execute(cmd)
+      puts cmd
+
+      begin
+      stdin, stdout, stderr = Open3.popen3(cmd)
+      result = stdout.gets
+      result.chomp! unless result.nil?
+      [result, stderr.gets]
+      rescue Exception => e
```
djydewang: I'm not sure if we're going to handle exception this way.

adangel: Were would the exception be raised? By the Open3.popen3 call? I guess, that means, that PMD either wasn't executed at all or not correctly, so we can't trust the result (if there is any).
Just printing the exception is fine for now. We might want to add a more speaking message/warning, such as "PMD execution failed. No results. Reason: " + e.message.

lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -0,0 +1,81 @@
...
+        end
+    end
+
+    def get_pmd_binary_file
+      Dir.chdir(@local_git_repo)
+
+      checkout_cmd = "git checkout " + @pmd_branch_name
+      PmdTester::Cmd.execute(checkout_cmd)
+
+      package_cmd = "./mvnw clean package -Dpmd.test.skip=true -Dpmd.skip=true"
```
adangel: You could also add `-Dmaven.test.skip=true` to skip unit test compiling and executing, too.

lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -0,0 +1,81 @@
...
+      PmdTester::Cmd.execute(package_cmd)
+
+      version_cmd = "./mvnw -q -Dexec.executable=\"echo\" -Dexec.args='${project.version}' " +
+          "--non-recursive org.codehaus.mojo:exec-maven-plugin:1.5.0:exec | tail -1"
```
adangel: The "tail -1" part is not needed anymore. It was a problem of an older maven wrapper version, but since then, we already updated.
(the old maven wrapper output first a line issueing the wrapper version itself before executing the command...)

lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -0,0 +1,81 @@
...
+      PmdTester::Cmd.execute(unzip_cmd)
+
+      Dir.chdir(@pwd)
+    end
+
+    def generate_pmd_report(src_root_dir, report_file)
+      run_path = "target/pmd-bin-#{@pmd_version}/bin/run.sh"
+      pmd_cmd = run_path + " -d " + src_root_dir + " -f xml -R" +  @rule_sets + " -r " + report_file
+      PmdTester::Cmd.execute(pmd_cmd)
+      if $?.to_i == 1
```
adangel: Maybe this check here can be moved into Cmd.execute? It would the raise an exception.....

test/test_cmd.rb
```diff
@@ -0,0 +1,9 @@
+require 'test/unit'
+require_relative '../lib/pmdtester/cmd'
+class TestCmd < Test::Unit::TestCase
+  def test_get_stdout
+    stdout, stderr = PmdTester::Cmd.execute("echo Hello, World!")
+    assert_equal("Hello, World!", stdout);
+    assert_nil(stderr)
+  end
+end
```
adangel: If you want add test cases to check the exit code of the command, you can use e.g. `/bin/false` to have a failing command....

lib/pmdtester/cmd.rb
djydewang: > Were would the exception be raised? By the Open3.popen3 call?

Refer to the [link](https://stackoverflow.com/questions/26040249/why-does-open3-popen3-return-wrong-error-when-executable-is-missing),  if we add a semicolon after CMD, then we don't need to think about the exception of Open3...
This has been bothering me for a long time, and now it's settled:)

lib/pmdtester/cmd.rb
```diff
@@ -0,0 +1,25 @@
...
+  class Cmd
+
+    PMD_EXIT_STATUS = 4
+
+    def self.execute(cmd)
+      puts cmd
+
+      stdin, stdout, stderr, wait_thr = Open3.popen3("#{cmd};")
+
+      unless wait_thr.value.success? || wait_thr.value.exitstatus == PMD_EXIT_STATUS
```
adangel: Do I understand this right? We consider here exit code 0 and exit code 4 as good (for the PmdTester). If PMD exits with any other code, we fail and stop any processing, correct? That is the right behavior.

Maybe rename "PMD_EXIT_STATUS" to "PMD_VIOLATIONS_FOUND".

djydewang: > Do I understand this right?

Yes, you are right. Hm... I have no good idea about how to add test cases for `PmdReportBuilder` now. Maybe we could merge the PR first.

adangel: I've just merged this PR.

We would ignore the exit code "4" for any command, regardless whether we executed PMD or some other command. This is not so good.

Since we are not interested in failing, when violations are found, we can use the command line option for pmd `-failOnViolation false` (https://pmd.github.io/pmd/pmd_userdocs_getting_started.html#exit-status) when we execute pmd.

About the test - yes. Definitely separate PR.
We could add/commit a small test repo under test resources (e.g. create a git init --bare repo) and use this for the testing.... then we don't depend on external resources for the unit tests.

lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -0,0 +1,94 @@
...
+      create_repositories_dir
+
+      @projects.each do |project|
+        path = @repositories_dir + project.name
+        clone_cmd = "#{project.type} clone #{project.connection} #{path}"
+
+        Cmd.execute(clone_cmd) unless File::exist?(path)
+
+        execute_reset_cmd(project.type, project.tag)
+        end
```
jsotuyod: this line seems to be indented more than necessary

lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    def get_projects
+      puts 'Cloning projects started'
+
+      create_repositories_dir
+
+      @projects.each do |project|
+        path = @repositories_dir + project.name
+        clone_cmd = "#{project.type} clone #{project.connection} #{path}"
```
jsotuyod: this doesn't support having the repository already cloned (ie: having it cached under Travis), which I forsee is going to happen... I'd at least have a TODO to remind ourselves later

lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -0,0 +1,94 @@
...
+        package_cmd = './mvnw clean package -Dpmd.test.skip=true -Dpmd.skip=true -Dmaven.test.skip=true'
+        Cmd.execute(package_cmd)
+
+        version_cmd = "./mvnw -q -Dexec.executable=\"echo\" -Dexec.args='${project.version}' " +
+            "--non-recursive org.codehaus.mojo:exec-maven-plugin:1.5.0:exec"
+        @pmd_version = Cmd.execute(version_cmd)
+
+        target_dir = "#@pwd/target"
+        unzip_cmd = "unzip -qo pmd-dist/target/pmd-bin-#{@pmd_version}.zip -d #{target_dir}"
+      Cmd.execute(unzip_cmd)
```
jsotuyod: this line should be indented one extra level

jsotuyod:
@djydewang @adangel shall we consider having something such as [Rubocop](https://github.com/bbatsov/rubocop) setup so we may automatically find most of these errors?

adangel:
@jsotuyod Good idea with robocup! We'll add it to the todo list :smile:

lib/pmdtester/builders/pmd_report_builder.rb
adangel: It currently only clones, if the repository directory (`path`) doesn't exist yet - see the following lines on master (https://github.com/pmd/pmd-regression-tester/blob/master/lib/pmdtester/builders/pmd_report_builder.rb#L40-L43). However, this will probably see more tweaking in the future...

PR #4, opened 2018-05-19
Test PmdReportBuilder

djydewang:


lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -36,13 +35,17 @@ def get_projects
       create_repositories_dir
 
       @projects.each do |project|
-        path = @repositories_dir + project.name
+        path = "#@repositories_dir/#{project.name}"
         clone_cmd = "#{project.type} clone #{project.connection} #{path}"
 
         Cmd.execute(clone_cmd) unless File::exist?(path)
```
adangel: I would add some output (via `puts`) here, to know, what's going on, like "Skipping clone, project path #{path} already exists" or so. It helps to understand, what's going on.

lib/pmdtester/cmd.rb
```diff
@@ -3,14 +3,13 @@
 module PmdTester
   class Cmd
 
-    PMD_EXIT_STATUS = 4
-
     def self.execute(cmd)
       puts cmd
 
       stdin, stdout, stderr, wait_thr = Open3.popen3("#{cmd};")
```
adangel: I'm having some trouble with running the test TestPmdReportBuilder locally. It seems, that the processes block (e.g. maven or pmd itself) until the output stream/err stream is consumed before they exit.
This is maybe the reason, why this PR doesn't build on travis - there is no output for more than 10 minutes, and then the build is automatically aborted.

I've seen this bug report: https://bugs.ruby-lang.org/issues/9082
At the very end, they suggest to use `capture3` instead of `popen3`, which seems to be an easy solution.
Note: It has quite a different syntax (e.g. doesn't return the stdin pipe, returns directly the exit value...)

PR #5, opened 2018-05-20
Add DiffBuilder for PmdTester

djydewang:
When analyzing violation diffs, it is assumed that the violations of file in the report are stored in order according to 'beginline'.  When parsing error diffs, it is assumed that parsing a file may throw multiple errors e.g different rules throw different errors.
The test for DiffBuilder is blocked by https://github.com/pmd/pmd-regression-tester/pull/4.

### DIFF
--- a/lib/pmdtester/builders/diff_builder.rb
+++ b/lib/pmdtester/builders/diff_builder.rb
@@ -8,7 +8,17 @@ module PmdTester
 
     def build(base_report, patch_report)
       base_doc = Nokogiri::XML(File.read(base_report))
+      patch_doc = Nokogiri::XML(File.read(patch_report))
 
+      violations_diffs = build_violation_diffs(base_doc, patch_doc)
+      error_diffs = build_error_diffs(base_doc, patch_doc)
+
+      [violations_diffs, error_diffs]
+    end
+
+    def build_violation_diffs(base_doc, patch_doc)
+
+      # key:String => value:Violation Array
       diffs = Hash.new
 
       base_doc.xpath('//file').each do |file|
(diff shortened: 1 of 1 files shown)


lib/pmdtester/builders/diff_builder.rb
```diff
@@ -0,0 +1,202 @@
...
+      file.xpath('violation').each do |violation|
+        violations.push(PmdViolation.new(violation, branch))
+      end
+      [filename, violations]
+    end
+
+    def remove_duplicate(base_violations, patch_violations)
+      i, j = 0, 0
+      diff = []
+      while i < base_violations.size && j < patch_violations.size
```
adangel: Do we maybe need to sort the violations first? I'm not sure, in which order the different rules will be executed and end up in the report, so potentially, the base and patch violations could contain exactly the same violations, but in total different order.

Have you looked at https://docs.ruby-lang.org/en/2.0.0/Hash.html? I know from java, there are good support methods to deal with Hashes (or Sets/HashSet). Since you already use concrete objects (PmdViolation), it might be easier to do the following: Implement `hash` and `eql?` in PmdViolation. Then you need to figure out the duplicates: Dups = Base intersect Patch . The end result would be "Base union Patch" minus "Duplicates". That should give you the violations, that are either in base or in patch, but not in both....
Or in Java:

```
Set<PmdViolation> result = new HashSet<>();
Set<PmdViolation> base = ...;
Set<PmdViolation> patch = ...;

result.addAll(base);
result.addAll(patch);

Set<PmdViolation> intersection = new HashSet<>(base);
intersection.retainAll(patch);

result.removeAll(intersection);
```

(Caveat: I didn't try this out....)

lib/pmdtester/builders/diff_builder.rb
```diff
@@ -0,0 +1,202 @@
...
+          diffs[filename] = [PmdError.new(error, 'patch')]
+        end
+      end
+      diffs
+    end
+  end
+
+  class PmdError
+    #The pmd branch type, 'base' or 'patch'
+    attr_reader :id
```
adangel: I wouldn't name this "id" - see comment below. Maybe "branch" or "sourceBranch"?

PR #6, opened 2018-05-23
Change the package command for building PMD

djydewang:


lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -57,7 +57,8 @@ def get_pmd_binary_file
         checkout_cmd = "git checkout #@pmd_branch_name"
         Cmd.execute(checkout_cmd)
 
-        package_cmd = './mvnw clean package -Dpmd.test.skip=true -Dpmd.skip=true -Dmaven.test.skip=true'
+        package_cmd = './mvnw clean package -Dpmd.skip=true -Dmaven.test.skip=true' +
```
adangel: oh, there is a missing space between `-Dmaven.test.skip=true` and `-Dmaven.checkstyle.skip=true`...

PR #7, opened 2018-05-24
Add test cases for DiffBuilder

djydewang:


lib/pmdtester/builders/diff_builder.rb
```diff
@@ -7,25 +7,29 @@ class DiffBuilder
     # http://pmd.sourceforge.net/report_2_0_0.xsd
 
     def build(base_report, patch_report)
-      base_doc = Nokogiri::XML(File.read(base_report))
-      patch_doc = Nokogiri::XML(File.read(patch_report))
+      base_doc = Nokogiri::XML(File.read(base_report)).remove_namespaces!
```
adangel: What was the problem with namespaces here? Just want to make sure, we don't have something weird in our schema...

lib/pmdtester/builders/diff_builder.rb
```diff
@@ -7,25 +7,29 @@ class DiffBuilder
...
     def build_diffs(base_hash, patch_hash)
-      base_hash.merge(patch_hash) do |key, base_value, patch_value|
-        base_value | patch_value - base_value & patch_value
+      diffs = base_hash.merge(patch_hash) do |key, base_value, patch_value|
+        (base_value | patch_value) - (base_value & patch_value)
+      end
+
+      diffs.delete_if do |key, value|
+        value.empty?
       end
```
adangel: Hm... is it for sure, that this function always returns "diffs"? E.g. does it really return diffs, even if the computed hash contained empty values and thus "value.empty?" has been called? I don't see exactly, what the last executed statement of this function is and personally I would add an explicit "return diffs" as the last statement (or just "diffs") to make it clear, that this is the result of the function...

lib/pmdtester/builders/diff_builder.rb
```diff
@@ -141,7 +156,7 @@ def initialize(violation, branch)
     def eql?(other)
       @violation['beginline'].eql?(other.violation['beginline']) &&
```
adangel: does this work or shouldn't that be rather "@violation.at_xpath('beginline')"? Same then for the 'rule' attribute.

lib/pmdtester/builders/diff_builder.rb
djydewang: @adangel 
> Hm... is it for sure, that this function always returns "diffs"?

Yes, the function alwats returns "diffs". Actually, `value.empty?` is is not accessible outside each's block, so we don't have to worry about the return value

lib/pmdtester/builders/diff_builder.rb
djydewang: Using `at_xpath` in `@error.at_xpath('filename').text` because `filename` is a subelement of `error` element.
`beginline` and `rule` is a attribute of `violation` element, so we should use `[attributename]` to get the value of attribute.

lib/pmdtester/builders/diff_builder.rb
djydewang: Don't worry. We don't care about the namespace when parsing the report of PMD, so we can just remove it. Refer to https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Document#remove_namespaces%21-instance_method

lib/pmdtester/builders/diff_builder.rb
```diff
@@ -26,23 +27,38 @@ def build_diffs(base_hash, patch_hash)
...
+    def get_diffs_size(diffs_hash)
+      size = 0
+      diffs_hash.keys.each do |key|
+        size += diffs_hash[key].size
+      end
+      size
     end
 
     def get_violations_hash(doc, branch)
       # key:string => value:PmdViolation Array
```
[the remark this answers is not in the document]
djydewang: @adangel Done.

lib/pmdtester/builders/diff_builder.rb
```diff
@@ -63,27 +79,35 @@ def get_violations_in_file(file, branch)
...
     def get_errors_hash(doc, branch)
       errors_hash = {}
+      errors_size = 0
 
       doc.xpath('//error').each do |error|
         filename = error.at_xpath('filename').text
         pmd_error = PmdError.new(error, branch)
         if errors_hash.has_key?(filename)
           errors_hash[filename].push(pmd_error)
+          errors_size += 1
```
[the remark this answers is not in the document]
djydewang: Done.

lib/pmdtester/builders/diff_builder.rb
```diff
@@ -63,27 +79,35 @@ def get_violations_in_file(file, branch)
...
 
-      build_diffs(base_hash, patch_hash)
+      error_diffs = build_diffs(base_hash, patch_hash)
+      error_diffs_size = get_diffs_size(error_diffs)
+      report_diff.error_diffs = error_diffs
+      report_diff.error_diffs_size =error_diffs_size
     end
 
     def get_errors_hash(doc, branch)
       errors_hash = {}
```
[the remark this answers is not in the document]
djydewang: Done.

PR #10, opened 2018-05-30
Add bundler to manage dependency

djydewang:


.travis.yml
```diff
@@ -1,3 +1,3 @@
 language: ruby
 jdk: oraclejdk9
-install: gem install hoe nokogiri slop
+install: bundler install
```
adangel: On https://docs.travis-ci.com/user/languages/ruby/ are the default values for these steps.
They use `bundle install --jobs=3 --retry=3` - which does 3 downloads in parallel and 3 retries, if a download failed.

I think, we should remove this line altogether in our `.travis.yml` file, so that we get always the latest defaults from travis.

PR #11, opened 2018-06-01
Using rubocop to check code style of the project

djydewang:
- Found something cool in http://rubocop.readthedocs.io/en/latest/automated_code_review/ It's late at night. I haven't looked at them carefully yet.

- The cool feature named 'auto correct' in rubocop, which Auto-correct certain offenses. If PMD also has this feature it would be great

- I'v fixed some offenses e.g. LineLength. Remaining offenes are ignored by .rubocop_todo.yml. We can keep fixing offenses and update  .rubocop.yml and .rubocop_todo.yml.

- Adding a rake task named check_style to check ruby code style by rubocop.

Edit: I'll make traivs green tomorrow.


jsotuyod:
> The cool feature named 'auto correct' in rubocop, which Auto-correct certain offenses. If PMD also has this feature it would be great

PMD has an ongoing project to start supporting this ^_^

jsotuyod:
@djydewang Travis is still failing :(

lib/pmdtester/builders/html_report_builder.rb
```diff
@@ -1,17 +1,17 @@
 require 'nokogiri'
 
 module PmdTester
+  # Building html report according to the output of DiffBuilder
   class HtmlReportBuilder
-
-    NO_DIFFERENCES_MESSAGE = 'No differences found!'
-    CSS_SRC_DIR = 'resources/css'
+    NO_DIFFERENCES_MESSAGE = 'No differences found!'.freeze
```
adangel: Interesting, didn't know about `freeze`. But yes, these are constants.

lib/pmdtester/parsers/options.rb
```diff
@@ -32,15 +33,21 @@ def parse(argv)
...
         o.string '-c', '--config', 'path to the base and patch PMD configuration file'
-        o.string '-l', '--list-of-project', 'path to the file which contains the list of standard projects', required: true
-        o.string '-m', '--mode', "the mode of the tool: 'local', 'online' or 'single'\n" +
-            "\tsingle: Set this option to 'single' if your patch branch contains changes for any option that can't work on master/base branch\n" +
-            "\tonline: Set this option to 'online' if you want to download the PMD report of master/base branch rather than generating it locally\n" +
-            "\tlocal: Default option is 'local'"
+        o.string '-l', '--list-of-project',
+                 'path to the file which contains the list of standard projects', required: true
+        o.string '-m', '--mode', "the mode of the tool: 'local', 'online' or 'single'\n" \
+                                 "\tsingle: Set this option to 'single' if your patch branch " \
```
adangel: Hm.... this now isn't so readable as it was before. Can you try to use a [Here Document](http://ruby-doc.org/core-2.5.1/doc/syntax/literals_rdoc.html#label-Here+Documents)? I'm not sure, whether it would directly work as an argument or whether you would need to define a local variable before, but a here document might help in improving the readability of the help text for option "--mode".

lib/pmdtester/parsers/projects_parser.rb
```diff
@@ -14,22 +15,22 @@ def parse(list_file)
...
-      document.xpath("//project").each do |project|
+      document.xpath('//project').each do |project|
         projects.push(Project.new(project))
       end
       projects
     end
 
     def get_schema_file
-      "config/projectlist_1_0_0.xsd"
+      'config/projectlist_1_0_0.xsd'
```
adangel: We can make this a constant, too.

.travis.yml
```diff
@@ -1,2 +1,5 @@
 language: ruby
 jdk: oraclejdk9
+script:
+  - rake test
```
adangel: Ok, with #10 merged, we use now bundle.... in order to use the dependencies for rake, we need to call it that way:

```
script:
  - bundle exec rake test
  - bundle exec rake check_style
```

It seems, that travis automatically uses bundle exec by default, if no custom script is configured, but now we have to do it manually.
This *should* fix the build....

PR #12, opened 2018-06-05
Fix Metrics/BlockLength offenses

djydewang:
The cop information: http://rubocop.readthedocs.io/en/latest/cops_metrics/#metricsblocklength

Attention: the default value of BlockLength is 25 although default value of [MethodLength](http://rubocop.readthedocs.io/en/latest/cops_metrics/#metricsmethodlength) is 10, it's strange.


lib/pmdtester/builders/html_report_builder.rb
```diff
@@ -122,53 +118,54 @@ def build_violations_section(doc, violation_diffs)
...
-              doc.td do
-                doc.a(id: "A#{a_index}", href: "#A#{a_index}") do
-                  doc.text '#'
-                end
-              end
+    def build_violation_table_body(doc, key, value)
+      doc.tbody do
+        a_index = 1
+        value.each do |pmd_violation|
+          doc.tr(class: pmd_violation.branch == 'base' ? 'a' : 'b') do
```
adangel: How does it change, if you factor out this block into a own method "build_violation_table_row(a_index, pmd_violation)"?

lib/pmdtester/builders/html_report_builder.rb
```diff
@@ -122,55 +118,60 @@ def build_violations_section(doc, violation_diffs)
...
+      doc.tbody do
+        a_index = 1
+        value.each do |pmd_violation|
+          build_violation_table_raw(a_index, key, pmd_violation, doc)
+          a_index += 1
+        end
+      end
+    end
+
+    def build_violation_table_raw(a_index, key, pmd_violation, doc)
```
adangel: The name should be actually "row" (not "raw").

I'd put the doc parameter as first, since for the other methods, the document is always the first parameter.
"key" will be the second parameter then (e.g. similar to the method "build_violation_table_body", or "build_violation_table").

PR #14, opened 2018-06-06
Add Runner to PmdTester

djydewang:
Investigation about mock frameworks:

- [rspec-mocks](https://relishapp.com/rspec/rspec-mocks/docs) it is a test-double framework for **rspec** with support for method stubs, fakes, and message expectations on generated test-doubles and real objects alike. It may not well applied to our project which is not a spec project.
- [Minitest](http://docs.seattlerb.org/minitest/Minitest/Mock.html) It is a simple and clean framework, but it is weak compared with mocha.
- [mocha](http://gofreerange.com/mocha/docs/) Built-in support for Test::Unit which we are using to do unit tests. Its funtion also is enough for our project.

I'll add the unit test for Runner as long as we confirm which framework to use.
And I'll add the integration test for runner as long as https://github.com/pmd/pmd-regression-tester/pull/13 is merged.


bin/pmdtester
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env ruby
-require 'pmdtester/runner'
+require '../lib/pmdtester/runner'
```
adangel: Shouldn't this be require_relative?

lib/pmdtester/builders/html_report_builder.rb
```diff
@@ -19,6 +19,7 @@ def build(project, report_diff)
       index.puts html_report
       index.close
 
+      puts "Build difference report of #{project.name} successfully!"
```
adangel: Small typo: Build -> Built

lib/pmdtester/runner.rb
```diff
@@ -0,0 +1,51 @@
...
+      PmdReportBuilder
+        .new(@options.patch_config, @projects, @options.local_git_repo, @options.patch_branch)
+        .build
+
+      @projects.each do |project|
+        report_diffs = DiffBuilder.new.build(project.pmd_reports[@options.base_branch],
+                                             project.pmd_reports[@options.patch_branch])
+        HtmlReportBuilder.new.build(project, report_diffs)
+      end
+      puts 'Build all difference reports successfully!'
```
adangel: Build -> Built

lib/pmdtester/runner.rb
```diff
@@ -0,0 +1,51 @@
...
+    def run_local_mode
+      PmdReportBuilder
+        .new(@options.base_config, @projects, @options.local_git_repo, @options.base_branch)
+        .build
+      PmdReportBuilder
+        .new(@options.patch_config, @projects, @options.local_git_repo, @options.patch_branch)
+        .build
+
+      @projects.each do |project|
+        report_diffs = DiffBuilder.new.build(project.pmd_reports[@options.base_branch],
```
adangel: Maybe we can add here a output, too - so that we see, what the PmdTester is currently doing. E.g. "Preparing report for #{project.name}". We can probably do much fine tuning in this area later on, though...

adangel:
Thanks for the list of possible frameworks - I'll have a look at them and get back to you.

test/integration_test_runner.rb
```diff
@@ -0,0 +1,15 @@
...
+  def test_runner
+    Process.fork do
+      argv = %w[-r target/repositories/pmd -b master -bc config/design.xml
+                -p pmd_releases/6.1.0 -pc config/design.xml -l test/resources/project-test.xml]
+
+      PmdTester::Runner.new(argv).run
+    end
+    Process.wait
+
+    assert_equal(0, $CHILD_STATUS.exitstatus)
```
adangel: That's already very good.
We should additional check at least for the existence of the generated reports - so, we expect a diff report html file to be written.

PR #16, opened 2018-06-11
Add single mode, add mocha library for unit test

djydewang:


Rakefile
```diff
@@ -11,6 +11,7 @@ Hoe.spec 'pmdtester' do
 
   self.author  = 'Binguo Bao'
   self.email   = 'djydewang@gmail.com'
+  self.clean_globs = %w[target/reports/**/* target/test**/*]
```
adangel: You can simply clean the whole target directory, no need to be specific here :smile:

lib/pmdtester/builders/diff_builder.rb
```diff
@@ -17,6 +17,27 @@ def build(base_report, patch_report)
       report_diff
     end
 
+    # Serving single mode
+    def build_single(patch_report)
+      report_diff = ReportDiff.new
+      base_doc = Nokogiri::XML('')
+      patch_doc = Nokogiri::XML(File.read(patch_report)).remove_namespaces!
+
+      # violations_hash, violations_size = get_violations_hash(patch_doc, 'base')
```
adangel: Please remove these comments - it looks like an old implementation...

lib/pmdtester/builders/diff_builder.rb
```diff
@@ -17,6 +17,27 @@ def build(base_report, patch_report)
       report_diff
     end
 
+    # Serving single mode
+    def build_single(patch_report)
```
adangel: Hm.... build and build_single are almost identical. Have you thought about using a default parameter? like

```
def build(base_report = nil, patch_report)
```

Hm... looks like, we would need a condition then inside, depending on whether base_report is given or not...

lib/pmdtester/runner.rb
```diff
@@ -39,14 +45,28 @@ def run_local_mode
...
-      # TODO
+      puts "Mode: #{@options.mode}"
+      puts "Patch branch/tag name: #{@options.patch_branch}"
+      puts "Patch config path: #{@options.patch_config}"
+
+      PmdReportBuilder
+        .new(@options.patch_config, @projects, @options.local_git_repo, @options.patch_branch)
+        .build
+      @projects.each do |project|
+        report_diffs = DiffBuilder.new.build_single(project.pmd_reports[@options.patch_branch])
```
adangel: If the build_single and build methods are the same, you can extract the loop in run_local_mode and run_single_mode into a own method.

test/integration_test_runner.rb
```diff
@@ -18,4 +18,21 @@ def test_runner
     assert_path_exist('target/reports/diff/checkstyle/index.html')
     assert_path_exist('target/reports/diff/pmd/index.html')
   end
+
+  def test_single_mode
+    `rake clean`
```
adangel: For consistency reasons, I would add `rake clean` also for the other test - or better, in the `setup` method of the unit test.

test/test_runner.rb
```diff
@@ -0,0 +1,37 @@
...
+require_relative '../lib/pmdtester/builders/html_report_builder'
+require_relative '../lib/pmdtester/builders/pmd_report_builder'
+require_relative '../lib/pmdtester/runner'
+
+# Unit test class for PmdTester::Runner
+class TestRunner < Test::Unit::TestCase
+  include PmdTester
+  def test_local_mode
+    report_diff = ReportDiff.new
+    PmdReportBuilder.any_instance.stubs(:build).returns(nil).twice
```
adangel: I need to get used to this style :smile: The solution in java would look completely different (you would inject mocks into the Runner module) - but in scripting languages, you can redefine everything....

test/test_runner.rb
```diff
@@ -0,0 +1,37 @@
...
+    DiffBuilder.any_instance.stubs(:build).returns(report_diff).twice
```
adangel: Is it possible, to verify the arguments, with which the build method is called?

Rakefile
djydewang: Maybe we shouldn't clone standard projects to `target/repositories` again and again.

adangel: OK, I understand. Hm... this sound like a future improvement - we should make the path, to which we clone the projects, configurable. Currently it is [fixed](https://github.com/pmd/pmd-regression-tester/blob/4b6b53953048d6578b59e6cab7b01f6826275841/lib/pmdtester/builders/pmd_report_builder.rb#L17).

So, not cleaning the whole `target` directory means, we get a slight performance improvements for the second integration test...

test/test_runner.rb
```diff
@@ -7,7 +7,27 @@
 
 # Unit test class for PmdTester::Runner
 class TestRunner < Test::Unit::TestCase
+  def setup
+    `rake clean`
+  end
+
   include PmdTester
+  def test_option_miss(argv, expects)
```
adangel: I'd call this method something like `run_and_assert_error_messages`. I first mistakenly thought, this would be a test of its own...

test/integration_test_runner.rb
```diff
@@ -1,6 +1,10 @@
 require 'test/unit'
 require_relative '../lib/pmdtester/runner'
 class IntegrationTestRunner < Test::Unit::TestCase
+  def setup
+    `rake clean`
+  end
+
   def test_local_mode
     Process.fork do
```
adangel: I think, we should change the integration tests, to execute the shell script `bin/pmdtester`, so that we have this covered in the tests, too.

You might need to call it again via bundler, e.g. `bundle exec bin/pmdtester -r target/repositories/pmd ...`.

adangel:
Merged! :tada:

djydewang:
@adangel 
> make the path, to which we clone the repositories configurable

If we make repositories path configurable without doing the same thing with pmd reports path, diff reports path etc.. it's strange. Maybe we can make `target` path configurable since PmdTester output everything to `target ` directory.

PR #17, opened 2018-06-16
Add more details about pmd branchs and pmd reports

djydewang:
Details about the pmd branch:

- Branch configuration
- Bnrach name
- Last commit SHA1 of the branch
- Last commit message of the branch
- The branch's execution time on all standard projects

Details about the pmd report:
- Execution time
- Timestamp of  the report

@adangel 
Maybe we can add more details to the diff report.
Please share your ideas about the details :smile:

TODO: details about the project list


test/test_runner.rb
```diff
@@ -28,22 +28,6 @@ def run_and_assert_error_messages(argv, expects)
...
-              -p pmd_releases/6.1.0 -pc config/design.xml -l test/resources/project-test.xml]
-    Runner.new(argv).run
-
-    actual_file_content = File.open('target/reports/diff/checkstyle/index.html').read
-    expected_file_content =
-      File.open('test/resources/html_report_builder/expected_empty_diff_report.html').read
-
-    assert_equal(expected_file_content, actual_file_content)
-  end
-
```
djydewang: This test case actually test the `HtmlReportBuilder` rather than `Runner`, so I remove it.

lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -88,17 +108,44 @@ def generate_pmd_reports
...
+
+  # This class represents all details about report of pmd
+  class PmdReportDetail
+    attr_reader :file_path
+    attr_reader :execution_time
+
+    def initialize(file_path, execution_time, time_stamp)
+      @file_path = file_path
+      @execution_time = execution_time
+      @time_stamp = time_stamp
```
adangel: time_stamp attribute is missing

lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -88,17 +108,44 @@ def generate_pmd_reports
...
     end
 
     def build
       get_projects
       generate_pmd_reports
     end
   end
+
+  # This class represents all details about branch of pmd
+  class PmdBranchDetail
```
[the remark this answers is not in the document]
adangel: I understand. There are a few advantages though, if you keep each class in a separate file:

* Navigation within text editors (like atom, ...) - if you want to open the source of a class, you know, that you need to open the file with the same name (or according to conventions in this case pmd_branch_detail.rb).
* The class/file pmd_report_builder doesn't grow too big - smaller modules, that serve one purpose are easier to comprehend

Right now, the classes are purely data holders. However, they could be more: If they now the local_git_repo directory, they could determine the last git sha and message on their own - moving the methods from pmd_report_builder to pmd_branch_detail.

PmdReportDetail is a pure data class, too. Although I don't see similar obvious functions, that would fit into it...

test/test_runner.rb
adangel: :+1: Yes, it's this test case: https://github.com/pmd/pmd-regression-tester/blob/568ed1c42f3acfc0468a1756a7fe115c28a56254/test/test_html_report_builder.rb#L60

test/test_runner.rb
```diff
@@ -68,17 +52,6 @@ def test_local_miss_patch_config
     run_and_assert_error_messages(argv, expects)
   end
 
-  def test_single_mode
```
adangel: Why did you remove this test case? We don't have specific asserts, yes, but we verify the interactions between Runner and ReportBuilder, DiffBuilder, HtmlReportBuilder, e.g. DiffBuilder needs to be called twice.

test/test_runner.rb
adangel: Actually, thinking this over: If we remove this test case, then we don't have a test, anymore, that runs the "happy path" - we are left with only error cases. We probably don't need to assert here the actual created files, but the important part is the interaction between Runner, PmdReportBuilder and DiffBuilder. And shouldn't there also be a HtmlBuilder involved?

djydewang: > but the important part is the interaction between Runner, PmdReportBuilder and DiffBuilder.

This make sence, I'll re-add the test case.

lib/pmdtester/pmd_branch_detail.rb
```diff
@@ -0,0 +1,35 @@
...
+module PmdTester
+  # This class represents all details about branch of pmd
+  class PmdBranchDetail
+    attr_accessor :branch_last_sha
+    attr_accessor :branch_last_message
+    attr_accessor :branch_name
+    # The branch's execution time on all standard projects
+    attr_accessor :execution_time
+
+    def load
```
adangel: Nice idea with load and save method!

test/test_pmd_branch_detail.rb
```diff
@@ -0,0 +1,23 @@
+require 'test/unit'
+require_relative '../lib/pmdtester/pmd_branch_detail'
+
+# Unit test class for PmdTester::Cmd
```
adangel: It's actually PmdTester::PmdBranchDetail :smile:

test/test_pmd_report_detail.rb
```diff
@@ -0,0 +1,17 @@
+require 'test/unit'
+require_relative '../lib/pmdtester/pmd_report_detail'
+
+# Unit test class for PmdTester::Cmd
```
adangel: PmdTester::PmdReportDetail

lib/pmdtester/pmd_report_detail.rb
```diff
@@ -0,0 +1,23 @@
+require 'json'
+
+module PmdTester
+  # This class represents all details about report of pmd
+  class PmdReportDetail
+    attr_reader :execution_time
+    attr_reader :time_stamp
+
+    def self.save(report_info_path, execution_time, time_stamp)
```
adangel: This looks a bit asymmetric: the save method is a class method, but the load method is a instance method.
I'd make the save method a instance method, too (with just one argument: report_info_path) and use the attributes instead of arguments.

PR #18, opened 2018-06-21
Add SummaryReportBuilder to PmdTester

djydewang:
The summary report [demo](https://djydewang.github.io/summary).

### DIFF
--- a/lib/pmdtester/builders/diff_report_builder.rb
+++ b/lib/pmdtester/builders/diff_report_builder.rb
@@ -7,6 +7,7 @@ module PmdTester
     NO_DIFFERENCES_MESSAGE = 'No differences found!'.freeze
 
     def build(project, report_diff)
+      project.diffs_exist = report_diff.diffs_exist?
       @project = project
       @report_diff = report_diff
 
--- a/lib/pmdtester/builders/diff_report_builder.rb
+++ b/lib/pmdtester/builders/diff_report_builder.rb
@@ -1,0 +1,219 @@
+require 'nokogiri'
+require_relative './html_report_builder'
+
+module PmdTester
+  # Building diff report according to the output of DiffBuilder
+  class DiffReportBuilder < HtmlReportBuilder
+    NO_DIFFERENCES_MESSAGE = 'No differences found!'.freeze
+
+    def build(project, report_diff)
+      @project = project
+      @report_diff = report_diff
+
+      index = File.new(project.diff_report_index_path, 'w')
+
+      html_report = build_html_report('pmd xml difference report')
+      copy_css(project.target_diff_report_path)
+
+      index.puts html_report
+      index.close
+
+      puts "Built difference report of #{project.name} successfully!"
+    end
+
+    def build_body(doc)
+      violation_diffs = @report_diff.violation_diffs
+      error_diffs = @report_diff.error_diffs
+      doc.body(class: 'composite') do
+        doc.div(id: 'contentBox') do
+          build_summary_section(doc)
+          build_violations_section(doc, violation_diffs)
+          build_errors_section(doc, error_diffs)
+        end
+      end
+    end
+
+    def build_summary_section(doc)
+      doc.div(class: 'section', id: 'Summary') do
+        doc.h2 'Summary:'
+        build_summary_table(doc)
+      end
+    end
+
+    def build_summary_table(doc)
+      doc.table(class: 'bodyTable', border: '0') do
+        doc.thead do
+          doc.tr do
+            doc.th 'Item'
+            doc.th 'Base'
+            doc.th 'Patch'
+            doc.th 'Difference'
+          end
+        end
+
+        build_summary_table_body(doc)
+      end
+    end
+
+    def build_summary_table_body(doc)
+      doc.tbody do
+        build_summary_row(doc, 'number of errors', @report_diff.base_errors_size,
+                          @report_diff.patch_errors_size, @report_diff.error_diffs_size)
+        build_summary_row(doc, 'number of violations', @report_diff.base_violations_size,
+                          @report_diff.patch_violations_size, @report_diff.violation_diffs_size)
+        build_summary_row(doc, 'execution time', @report_diff.base_execution_time,
+                          @report_diff.patch_execution_time, @report_diff.diff_execution_time)
+        build_summary_row(doc, 'timestamp', @report_diff.base_timestamp,
+                          @report_diff.patch_timestamp, '')
+      end
+    end
+
+    def build_summary_row(doc, item, base, patch, diff)
+      doc.tr do
+        doc.td(class: 'c') { doc.text item }
+        doc.td(class: 'a') { doc.text base }
+        doc.td(class: 'b') { doc.text patch }
+        doc.td(class: 'c') { doc.text diff }
+      end
+    end
+
+    def build_filename_h3(doc, filename)
+      doc.h3 do
+        doc.a(href: @project.get_webview_url(filename)) do
+          doc.text @project.get_path_inside_project(filename)
+        end
+      end
+    end
+
+    def build_violations_section(doc, violation_diffs)
+      doc.div(class: 'section', id: 'Violations') do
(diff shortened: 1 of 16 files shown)


lib/pmdtester/builders/diff_report_builder.rb
```diff
@@ -0,0 +1,220 @@
+require 'nokogiri'
+require_relative './html_report_builder'
+
+module PmdTester
+  # Building diff report according to the output of DiffBuilder
+  class DiffReportBuilder < HtmlReportBuilder
+    NO_DIFFERENCES_MESSAGE = 'No differences found!'.freeze
+
+    def build(project, report_diff)
+      project.diffs_exist = report_diff.diffs_exist?
```
adangel: This looks a bit strange: Is this the correct place, to handover this information?
Maybe it's better done, before the diffreportbuilder is called?

lib/pmdtester/builders/summary_report_builder.rb
```diff
@@ -0,0 +1,146 @@
...
+    def build(projects, base_name, patch_name)
+      @projects = projects
+      @base_details = get_branch_details(base_name)
+      @patch_details = get_branch_details(patch_name)
+
+      FileUtils.mkdir_p(REPORT_DIR) unless File.directory?(REPORT_DIR)
+      index = File.new(INDEX_PATH, 'w')
+
+      html_report = build_html_report('Summary report')
+      copy_css(REPORT_DIR)
```
adangel: I think, this should be optimized. We copy the css files for the summary report and for each project report.
We should be able to reuse the css files for the project reports from the summary report.

lib/pmdtester/builders/summary_report_builder.rb
```diff
@@ -0,0 +1,146 @@
...
+      details
+    end
+
+    def build_body(doc)
+      build_branch_details_section(doc)
+      build_projects_section(doc)
+    end
+
+    def build_branch_details_section(doc)
+      doc.div(class: 'section', name: 'Branch details') do
```
adangel: The name attribute should be the id. And just a personal preference of me: I tend to use only lower-case letters for the ids and avoid using any special characters and spaces.

lib/pmdtester/builders/summary_report_builder.rb
```diff
@@ -0,0 +1,146 @@
...
+      doc.thead do
+        doc.tr do
+          doc.th 'Item'
+          doc.th 'base'
+          doc.th 'patch'
+        end
+      end
+    end
+
+    def build_branch_table_body(doc)
```
adangel: I just got an idea, what could be useful: github allows to craft urls, that shows you the diff between two commits/branches, e.g. https://github.com/pmd/pmd/compare/master...adangel:issue-1186
So, in general: https://github.com/pmd/pmd/compare/#{base_branch}...#{patch_branch}
On the other hand, when we go on to integrate this in pull request builds, we probably can provide a link to the original pull request.

lib/pmdtester/builders/summary_report_builder.rb
```diff
@@ -0,0 +1,146 @@
...
+    def build_branch_table_row(doc, item, base, patch)
+      doc.tr do
+        doc.td(class: 'c') { doc.text item }
+        doc.td(class: 'a') { doc.text base }
+        doc.td(class: 'b') { doc.text patch }
+      end
+    end
+
+    def build_projects_section(doc)
+      doc.div(class: 'section', name: 'Projects') do
```
adangel: name -> id

lib/pmdtester/builders/summary_report_builder.rb
```diff
@@ -0,0 +1,146 @@
...
+        build_projects_table_head(doc)
+        build_projects_table_body(doc)
+      end
+    end
+
+    def build_projects_table_head(doc)
+      doc.thead do
+        doc.tr do
+          doc.th 'project diff report link'
+          doc.th 'project name'
```
adangel: I would actually remove the first column and make the project name the link to the reports.

test/test_summary_report_builder.rb
```diff
@@ -0,0 +1,20 @@
+require 'test/unit'
+require_relative '../lib/pmdtester/builders/summary_report_builder'
+require_relative '../lib/pmdtester/parsers/projects_parser'
+
+# Unit test class for PmdTester::SummaryReportBuilder
+class TestSummaryReportBuilder < Test::Unit::TestCase
+  def test_summary_report_builder
```
adangel: We probably need here a `rake clean` call, too.

lib/pmdtester/builders/diff_report_builder.rb
adangel: Ok, I've now seen, where it is used. First, for each project, the html report is created - and this sets the flag "diffs_exist" on the project. The flag is later on used for the summary report. So it is crucial, to render the reports of each project into html before rendering the summary... this is a flow dependency.

How about using the project object for storing the results (the report_diff object), too? That way, the summary builder can ask the project, whether a diff exists and the project in turn asks the report diff.
The project object would start out with the basic information from the project list (the configuration with which PmdTester has been started) and we would add more information with the progress.
If you rework this, I suggest to do this in a new PR.

adangel:
@djydewang I've updated the wiki page with an example: https://github.com/pmd/pmd/wiki/%5BGSoC-2018%5D-Automated-regression-tests-against-real-world-projects#example

I've now also an account at chunk.io
here's one real report: https://chunk.io/adangel/2e9e4bc8f6b840459e3dfc1770d68ed6/index.html

I've noticed the following:

There seems to be a problem with last commit sha of the patch (it is the commit message... maybe, this happens if the patch is a git tag?).

I also noticed, that the execution time of the single projects is much better readable (00:03:31 vs. 262.89510442).

The diff exist flag seems to be not correct - it shows false on the summary, but the projects actually have some diffs...

lib/pmdtester/builders/diff_report_builder.rb
```diff
@@ -0,0 +1,220 @@
+require 'nokogiri'
+require_relative './html_report_builder'
+
+module PmdTester
+  # Building diff report according to the output of DiffBuilder
```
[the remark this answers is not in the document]
djydewang: Done.

lib/pmdtester/builders/summary_report_builder.rb
djydewang: Each project has its own css resource file that can make each project's diff html report independent,
which means we can upload single project diff report to somewhere, although, I don't know if there will be such a usage scenario in the future :smile:

lib/pmdtester/builders/summary_report_builder.rb
djydewang: Done.

lib/pmdtester/builders/summary_report_builder.rb
djydewang: Done.

lib/pmdtester/builders/summary_report_builder.rb
djydewang: Done.

test/test_summary_report_builder.rb
djydewang: Done.

djydewang:
> There seems to be a problem with last commit sha of the patch (it is the commit message... maybe, this happens if the patch is a git tag?).
> 
> I also noticed, that the execution time of the single projects is much better readable (00:03:31 vs. 262.89510442).
> 
> The diff exist flag seems to be not correct - it shows false on the summary, but the projects actually have some diffs...

I'v fixed issues above. And I've noticed the following:

- links to source code of open projects are wrong. the `webview url` should be `{connection_url}/tree/{branch/tag}` rather than just equal `connection url` for those projects that are stored on github. https://github.com/pmd/pmd-regression-tester/blob/e0c9ba80e0dc22b3a99e8903322e7dfb42eba7bf/lib/pmdtester/project.rb#L21
We can fixed it by providing` webview-url` for the project or generate right `webview-url` for project when they aren't provided.

PR #19, opened 2018-06-25
Add online mode for PmdTester

djydewang:
- Fix defual vlaue of webview url
- Add report_diff attribute to the Project class
- Finish online mode

I'v upload a baseline demo to [sourceforge](https://sourceforge.net/projects/pmd/files/pmd-regression-tester/) :smile:

### DIFF
--- a/lib/pmdtester/builders/diff_builder.rb
+++ b/lib/pmdtester/builders/diff_builder.rb
@@ -9,7 +9,7 @@ module PmdTester
     def build(base_report, patch_report, base_info, patch_info)
       # Serving for 'single' mode,
       # if value of `base_report` is nil then `base_doc` is empty
-      base_doc = if File.exist?(base_report)
+      base_doc = if !base_report.nil?
                    Nokogiri::XML(File.read(base_report)).remove_namespaces!
                  else
                    Nokogiri::XML('')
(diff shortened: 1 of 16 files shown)


lib/pmdtester/runner.rb
```diff
@@ -49,19 +51,63 @@ def run_local_mode
...
+      get_projects(@options.project_list) unless @options.nil?
+      branch_details = PmdReportBuilder
+                       .new(@options.patch_config, @projects,
+                            @options.local_git_repo, @options.patch_branch)
+                       .build
+      # copy list of projects file to the patch baseline
+      FileUtils.cp(@options.project_list, branch_details.target_branch_project_list_path)
 
-      build_html_reports
+      build_html_reports unless @options.html_flag
```
adangel: I'm thinking about your new html flag.... and why we need this...
I guess, if we can only call build_html_reports if we have both a base and patch report available. So, if we run single mode with a freshly installed tool (nothing in target yet), then we have no base report and thus can't create the diff report.

Shouldn't we just remove this flag and just never execute build_html_reports when running in single mode?

djydewang: Single mode is not just for baseline in my imagination.
It can also be used to evaluate the performance of new rules. For example, patch config contains only one new rule. Then we can see how this new rule behaves on all standard projects.

lib/pmdtester/builders/diff_builder.rb
```diff
@@ -27,9 +27,9 @@ def build(base_report, patch_report, base_info, patch_info)
 
     def build_detail_diffs(base_info, patch_info, report_diff)
       base_details = PmdReportDetail.new
-      base_details.load(base_info)
+      base_details.load(base_info) unless base_info.nil?
```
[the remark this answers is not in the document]
djydewang: > we should enforce/encourage to have always a base_info/patch_info file available 

Yes, the ideal situation is this. But in single mode, base_info file is not available, so we can't load it.
Also see the https://github.com/pmd/pmd-regression-tester/pull/19#discussion_r198741153

lib/pmdtester/runner.rb
adangel: Understood, yes, that makes sense.

PR #20, opened 2018-06-29
Change the way of parsing xml file from DOM to SAX

djydewang:


lib/pmdtester/parsers/pmd_report_document.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    def characters(string)
+      @current_violation.text = string unless @current_violation.nil?
+      @current_error.text = string unless @current_error.nil?
+    end
+
+    def end_element(name)
+      case name
+      when 'file'
+        @violations.violations.store(@current_filename, @current_violations)
```
adangel: From object-oriented design, this violates the principle of encapsulation. E.g. PmdReportDocument now needs to know the inner structure of PmdViolations (that is stores the violations in a hash filename -> array of violations). You can create a method in PmdViolations, such that this line reads:

```
@violations.add_violations_by_filename(@current_filename, @current_violations)
```

lib/pmdtester/parsers/pmd_report_document.rb
```diff
@@ -0,0 +1,65 @@
...
+        @current_filename = nil
+      when 'violation'
+        @current_violations.push(@current_violation)
+        @current_violation = nil
+      when 'error'
+        if @errors.errors.key?(@current_filename)
```
adangel: That's the same: You can put this logic into PmdErrors, that you don't need to bother here, how the errors are actually stored. It would be here just:

```
@errors.add_error_by_filename(@current_filename, @current_error)
```

and this method in PmdErrors takes care of the rest (either adding it to the existing array or put a new array into the hash).

lib/pmdtester/parsers/pmd_report_document.rb
```diff
@@ -0,0 +1,65 @@
...
+          @errors.errors[@current_filename].push(@current_error)
+        else
+          @errors.errors.store(@current_filename, [@current_error])
+        end
+        @current_filename = nil
+        @current_error = nil
+      when 'pmd'
+        @violations.violations_size = @violations_size
```
adangel: I'm wondering, whether PmdViolations/PmdErrors shouldn't be able to determine the total number of violations/errors by themselves? When you add the methods add_violations_by_filename and add_error_by_filename, you can do this.

However, keeping track of the number of violations separately seems to be an early optimization. It avoids some loops through the hashes to count the violations later on, but how often do we need this count?
At least, I'd move this house-holding task into PmdViolations/PmdErrors and provide a method to get the total number of violations/errors.

lib/pmdtester/parsers/pmd_report_document.rb
```diff
@@ -0,0 +1,65 @@
...
+      @branch_name = branch_name
+    end
+
+    def start_element(name, attrs = [])
+      @current_element = name
+
+      case name
+      when 'file'
+        @current_violations = []
+        @current_filename = attrs[0][1]
```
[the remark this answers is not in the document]
adangel: I've seen now, as you used it later on: `attrs.to_h` looks like an easy solution.
You can actually convert the attrs array to a hash here directly in this method and pass on the hash to the PmdViolation/PmdError constructors.

test/integration_test_runner.rb
```diff
@@ -7,10 +7,10 @@ def setup
   end
 
   def test_local_mode
-    argv = '-r target/repositories/pmd -b master -bc config/design.xml ' \
-            '-p pmd_releases/6.1.0 -pc config/design.xml -l test/resources/project-test.xml'
+    argv = %w[-r target/repositories/pmd -b master -bc config/design.xml
+              -p pmd_releases/6.1.0 -pc config/design.xml -l test/resources/project-test.xml]
 
-    Cmd.execute("bundle exec bin/pmdtester #{argv}")
```
adangel: Hm... The idea behind calling the shell script is, that we actually test the shell script, too. Otherwise, we don't have a integration test for the shell script, and we don't see, when we break the shell script...
We should have at least one test, that calls the PmdTester via the shell script. You could also create a separate integration test ruby script (like `integration_test_pmdtester.rb`).

PR #21, opened 2018-07-04
Add auto-gen-config option for PmdTester

djydewang:
TODO: Filter Pmd errors in the baseline based on dynamic configuration. 
Proposed method: check whether ruleset name(e.g 'design') is substring of pmd errors' details


lib/pmdtester/runner.rb
```diff
@@ -58,15 +61,19 @@ def run_online_mode
 
       baseline_path = download_baseline(@options.base_branch)
 
-      # patch branch build pmd reports with same configuration as base branch
-      config_path = "#{baseline_path}/config.xml"
+      if @options.auto_config_flag
```
djydewang: Only online mode support auto-gen-config option now, maybe we should also extend local mode.

lib/pmdtester/builders/rule_set_builder.rb
```diff
@@ -0,0 +1,107 @@
...
+      rule_sets = get_rule_sets(filenames)
+      build_config_file(rule_sets)
+    end
+
+    def filter_set(rule_sets)
+      if rule_sets.size.eql?(ALL_RULE_SETS.size)
+        # if `rule_sets` contains all rule sets, than no need to filter the baseline
+        @options.filter_set = nil
+      else
+        @options = rule_sets
```
adangel: That makes me wonder, what type `@options` is... is it a Set, a Object, or something else?
Note: I've not seen all the code in the PR yet, so it's difficult to grasp without seeing, how it is used.
But, if you assign here rule_sets (supposedly a Set) to `@options` means: It's a Set. But two lines above, you set a attribute `filter_set`... so `@options` can't be a Set...

adangel: Maybe this should be `@options.filter_set = rule_sets`?

lib/pmdtester/builders/rule_set_builder.rb
```diff
@@ -0,0 +1,107 @@
...
+      end
+    end
+
+    def diff_filenames
+      filenames = nil
+      Dir.chdir(@options.local_git_repo) do
+        diff_cmd = "git diff --name-only #{@options.base_branch}..#{@options.patch_branch}"
```
adangel: Do me a favor and add a little comment, that we only need to support git here, since PMD's repo is using git...
I caught myself thinking about the need for mercurial etc... :smile:

lib/pmdtester/builders/rule_set_builder.rb
```diff
@@ -0,0 +1,107 @@
...
+        grep_cmd = "grep -E '^pmd-(core|java)/'" # OPTIMIZE: extend other languages
```
adangel: Maybe you can combine these two commands into one git diff line:

`git diff --name only #{@options.base_branch}..#{@options.patch_branch} -- pmd-core/ pmd-java/`

I've not tried it, but the manual page of git diff suggests, that this should be possible.

For other languages, I think, there is a bit more to do, such as `PATH_TO_PMD_JAVA_BASE_RULES`...

lib/pmdtester/builders/rule_set_builder.rb
```diff
@@ -0,0 +1,107 @@
...
+
+require 'nokogiri'
+require 'set'
+require_relative '../cmd'
+module PmdTester
+  # This class is responsible for generation dynamic configuration
+  # according to the difference between base and patch branch of Pmd.
+  class RuleSetBuilder
+    ALL_RULE_SETS = Set['bestpractices', 'codestyle', 'design', 'documentation', 'errorprone',
+                        'migrating', 'multithreading', 'performance', 'regex', 'security'].freeze
```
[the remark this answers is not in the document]
adangel: Ok, these additionally must match the names of the xml ruleset files for the categories and the package names of the packages, in which the rules are... (in case of Java based rules).
I think, by convention, this works. I just wasn't aware about the (sub-)package name requirement, until I've seen the method `add_rule_set`.

lib/pmdtester/builders/rule_set_builder.rb
```diff
@@ -0,0 +1,107 @@
...
+        filenames = Cmd.execute("#{diff_cmd} | #{grep_cmd}")
+      end
+      filenames.split("\n")
+    end
+
+    def get_rule_sets(filenames)
+      rule_sets = Set[]
+      filenames.each do |filename|
+        if filename.start_with?(PATH_TO_PMD_JAVA_BASED_RULES, PATH_TO_PMD_XPATH_BASED_RULES)
+          add_rule_set(rule_sets, filename)
```
adangel: Instead of searching in the Set of known categories, you could also use a regex pattern to extract the category directly from the filenames.
There are two cases:
1. filename.start_with?(PATH_TO_PMD_JAVA_BASED_RULES)
`category = %r{pmd-java/src/main/java/net/sourceforge/pmd/lang/java/rule/([^/]+)/[^/]+Rule.java}.match(filename)[1]`

It defines one capturing group to extract the category from the directory name.

2. filename.start_with?(PATH_TO_PMD_XPATH_BASED_RULES)
`category = %{pmd-java/src/main/resources/category/java/([^/]+).xml}.match(filename)[1]`

This captures the filename without the extension.

See https://ruby-doc.org/core-2.1.1/Regexp.html#class-Regexp-label-Capturing

Note: If there is no match, then `match` returns nil. Keep that in mind, when you eliminate the start_with check - since this can be checked by the regex, too. either it matches (and a match returns a MatchData object) or it doesn't, then you need to do the second regex.

adangel: We could later on generalize these regex even more, e.g. extracting also the language... but not now.

lib/pmdtester/builders/rule_set_builder.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    def build_config_file(rule_sets)
+      if rule_sets.empty?
+        puts NO_JAVA_RULES_CHANGED_MESSAGE
+        exit 0
+      end
+
+      doc = Nokogiri::XML(File.read(PATH_TO_ALL_JAVA_RULES))
+      doc.search('rule').each do |rule|
+        rule.remove unless match_ref?(rule, rule_sets)
```
adangel: Since we are dealing now on the level of complete rulesets, you could easily generate out a new ruleset referencing only the changes categories:

Pseudo code:
```
rule_sets.each do |rule_set|
    puts "<rule ref=\"category/java/#{rule_set}.xml\">"
end
```

We can reference the complete rulesets/categories. You'll need of course the surrounding `<ruleset>` tag and the XML prolog. And maybe use a library to create the xml...

That way, we avoid looping through all the rules and repeatedly through the selected ruleset...

lib/pmdtester/builders/rule_set_builder.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    def write_dynamic_file(doc)
+      File.open(PATH_TO_DYNAMIC_CONFIG, 'w') do |x|
+        x << doc.to_s.gsub(/\n\s+\n/, "\n")
+      end
+      @options.base_config = PATH_TO_DYNAMIC_CONFIG
+      @options.patch_config = PATH_TO_DYNAMIC_CONFIG
+    end
+
+    def remove_empty_lines(doc)
```
adangel: Seems to be unused...

lib/pmdtester/parsers/pmd_report_document.rb
```diff
@@ -36,22 +39,37 @@ def start_element(name, attrs = [])
...
+
+    def cdata_block(string)
+      @current_error.text = string unless @current_error.nil?
+    end
+
+    def match_filter_set?(violation)
+      return true if @filter_set.nil?
+
+      @filter_set.each do |ruleset|
+        return true if ruleset.eql?(violation.attrs['ruleset'].delete(' ').downcase)
```
adangel: That's a pity, here ends the convention... right, the ruleset names have unfortunately camel case with spaces between the words... ("Best Practices" <-> bestpractices). Yes, we'll just have to "normalize" it here...

lib/pmdtester/runner.rb
adangel: :+1: sounds good

lib/pmdtester/builders/rule_set_builder.rb
adangel: Ok, didn't know that PATH_TO_ALL_JAVA_RULES already contains the references to the complete rulesets, so these two nested loops are not too big actually. Still I'd try to avoid too many loopings...

test/test_rule_set_builder.rb
```diff
@@ -0,0 +1,48 @@
...
+    diff_filenames = <<~DOC
+      pmd-java/src/main/java/net/sourceforge/pmd/lang/java/rule/design/NcssCountRule.java
+      pmd-java/src/main/java/net/sourceforge/pmd/lang/java/rule/codestyle/UnnecessaryReturnValue.java
+      pmd-java/src/main/java/net/sourceforge/pmd/lang/java/rule/codestyle/UnnecessaryConstructor.java
+    DOC
+    mock_build(diff_filenames)
+
+    expected = File.read("#{PATH_TO_TEST_RESOURCES}/expected-design-codestyle.xml")
+    actual = File.read(RuleSetBuilder::PATH_TO_DYNAMIC_CONFIG)
+    assert_equal(expected, actual)
```
adangel: We should also verify, that `options.filter_set` is correctly populated...

lib/pmdtester/builders/diff_report_builder.rb
```diff
@@ -1,10 +1,12 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
djydewang: Yes, refers to http://rubocop.readthedocs.io/en/latest/cops_style/#stylefrozenstringliteralcomment

lib/pmdtester/builders/rule_set_builder.rb
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
+
+require 'nokogiri'
+require 'set'
+require_relative '../cmd'
+module PmdTester
+  # This class is responsible for generation dynamic configuration
+  # according to the difference between base and patch branch of Pmd.
```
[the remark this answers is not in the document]
djydewang: Done.

lib/pmdtester/builders/rule_set_builder.rb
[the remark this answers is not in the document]
djydewang: I found that the `UnnecessaryCastRule` in migrating directory have not been referenced by any rule sets, so I removed 'migrating'. And I also remove 'regex' since [RegexHelper](https://github.com/pmd/pmd/blob/d6ec50b20f5a24dc8ceb083d792c41cebc930a35/pmd-java/src/main/java/net/sourceforge/pmd/lang/java/rule/regex/RegexHelper.java) in regex directory is not a rule class.

lib/pmdtester/builders/rule_set_builder.rb
djydewang: > Maybe this should be @options.filter_set = rule_sets?

Yes, it's my fault. I should add a test case for it.

lib/pmdtester/builders/rule_set_builder.rb
djydewang: Done :smile:

lib/pmdtester/builders/rule_set_builder.rb
djydewang: > I've not tried it, but the manual page of git diff suggests, that this should be possible.

It works well.

test/test_rule_set_builder.rb
djydewang: Done.

PR #23, opened 2018-07-08
Preparing for the release of PmdTester

djydewang:
Build the gem file pmdtester-1.0.0.pre.SNAPSHOT.gem: `bundle exec rake gem`
Install the package as a gem: `bundle exec rake install_gem`
TODO: 
* Add more test cases for the CLI
* Upload `pmdtester` gem to somewhere e.g. https://rubygems.org/


.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: ruby
-jdk: oraclejdk9
+jdk: openjdk10
 script:
   - bundle exec rake rubocop
   - bundle exec rake clean test
   - bundle exec rake clean integration-test
+  - bundle exec rake install_gem
+  - pmdtester -h
```
djydewang: I don't why CI failed when executed this command, though it works well locally.

Manifest.txt
```diff
@@ -1,8 +1,67 @@
...
+lib/pmdtester/pmd_error.rb
+lib/pmdtester/pmd_report_detail.rb
+lib/pmdtester/pmd_violation.rb
+lib/pmdtester/project.rb
+lib/pmdtester/report_diff.rb
+lib/pmdtester/resource_locator.rb
+lib/pmdtester/runner.rb
+resources/css/maven-base.css
+resources/css/maven-theme.css
+test/integration_test_pmd_report_builder.rb
```
adangel: Should we really include the tests in the package? I think, they are not necessary....

README.md
```diff
@@ -1,7 +1,9 @@
 # pmdtester
 
-code  :: https://github.com/pmd/pmd-regression-tester  
-bugs  :: https://github.com/pmd/pmd-regression-tester/issues  
+code  :: https://github.com/pmd/pmd-regression-tester
+
+bugs  :: https://github.com/pmd/pmd-regression-tester/issues
+
```
adangel: We can add here the badge from travis:

```
[![Build Status](https://travis-ci.com/pmd/pmd-regression-tester.svg?branch=master)](https://travis-ci.com/pmd/pmd-regression-tester)
```

It shows as: [Build Status](https://travis-ci.com/pmd/pmd-regression-tester)

adangel: I know now, what the double colons mean: It's not markdown, it's the ruby documentation format...
I guess, we should rename the file to be `README.rdoc`

The build status line looks then like this:

```
build-status :: {<img src="https://travis-ci.com/pmd/pmd-regression-tester.svg?branch=master" alt="Build Status" />}[https://travis-ci.com/pmd/pmd-regression-tester]
```

And remove the extra line you introduce between code and bugs. The double colon makes this a nice table, see e.g. https://github.com/ruby/rake/blob/master/README.rdoc

Note: the headings are also different, you can see the syntax here: https://raw.githubusercontent.com/ruby/rdoc/master/ExampleRDoc.rdoc

Manifest.txt
adangel: It seems, that rake doesn't include the test files.
Although on https://guides.rubygems.org/what-is-a-gem/ this is said:

> Inside gems are the following components:
> *    Code (including tests and supporting utilities)

I would actually remove the tests here. That way, if we add new tests, remove, rename, or so, we don't need to think about the package...

PR #24, opened 2018-07-10
Adding a logging framework for PmdTester

djydewang:
The PR did the following things:

- Add a option named 'debug' to PmdTester, let the user choose whether to output the log in debug mode.
- Add a global logger for PmdTester.


lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -27,21 +27,29 @@ def execute_reset_cmd(type, tag)
...
-      puts 'Cloning projects started'
+      logger.info 'Cloning projects started'
 
       @projects.each do |project|
         path = project.local_source_path
         clone_cmd = "#{project.type} clone #{project.connection} #{path}"
         if File.exist?(path)
-          puts "Skipping clone, project path #{path} already exists"
+          logger.warn "Skipping clone, project path #{path} already exists"
+          update_cmd = "#{project.type} pull #{project.connection}"
```
adangel: Good, you added the update :+1: 

Does this already work? For git, I think, we need to specify a branch, from which we pull? (since project.connection is a URL).
As we cloned it earlier, we could rely on the defaults and just do a "git pull origin".... which hopefully updates the master branch...
For mercurial, I'm not sure...

lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -27,21 +27,29 @@ def execute_reset_cmd(type, tag)
...
+
+          begin
+            Cmd.execute(update_cmd)
+          rescue CmdException
+            logger.warn "updating repository #{path} failed, remove the repository"
+            FileUtils.remove_dir(path)
```
adangel: Nice idea, but I'm not sure, if we really should automate this... 

+: It's self healing

-: It might be faster (more efficient), if you manually look into the error and fix it. Depending on the problem (e.g. currently no network connection), we would automatically delete the clone and you end up with nothing. And whoever uses pmdtester should be sure then, to never reuse any local repositories with the tool, since they might get deleted...

So, I'd vote for just exiting the tool in this case of error.

lib/pmdtester/cmd.rb
```diff
@@ -5,20 +5,36 @@
 module PmdTester
   # Containing the common method for executing shell command
   class Cmd
+    extend PmdTester
     def self.execute(cmd)
-      puts cmd
+      logger.info "execute command '#{cmd}'"
```
adangel: This should be debug. We probably have in the calling code already a log.info, explaining that we will execute some command (not the exact command line, but that's part of the debug logging then).

lib/pmdtester/builders/pmd_report_builder.rb
djydewang: I thought about it carefully, we shouldn't update the warehouse, we should choose the constant branch of the open source project.This will ensure that baselines are generated based on the same source code especially when it comes to online mode.

PR #25, opened 2018-07-12
Remove working directory substring from filename of pmd violation

djydewang:
When we generate the baseline which will be uploaded to somewhere, the file attribute of the violation element in the pmd report contains the local working path information. The baseline is used for sharing, so we should remove this part of the information.

### DIFF
--- a/lib/pmdtester/parsers/pmd_report_document.rb
+++ b/lib/pmdtester/parsers/pmd_report_document.rb
@@ -40,7 +40,7 @@ module PmdTester
     end
 
     def remove_work_dir!(str)
-      str.sub!(/^#{@working_dir}/, '')
+      str.sub!(/#{@working_dir}/, '')
     end
 
     def characters(string)
@@ -65,6 +65,7 @@ module PmdTester
     end
 
     def cdata_block(string)
+      remove_work_dir!(string)
       @current_error.text = string unless @current_error.nil?
     end
 
--- a/lib/pmdtester/parsers/pmd_report_document.rb
+++ b/lib/pmdtester/parsers/pmd_report_document.rb
@@ -9,7 +9,7 @@ module PmdTester
   class PmdReportDocument < Nokogiri::XML::SAX::Document
     attr_reader :violations
     attr_reader :errors
-    def initialize(branch_name, filter_set = nil)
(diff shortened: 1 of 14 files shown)


lib/pmdtester/parsers/pmd_report_document.rb
```diff
@@ -28,11 +29,11 @@ def start_element(name, attrs = [])
       case name
       when 'file'
         @current_violations = []
-        @current_filename = attrs['name']
+        @current_filename = attrs['name'].sub(/^#{@working_dir}/, '')
```
adangel: That's the hash key, that is used to collect the violations...

lib/pmdtester/parsers/pmd_report_document.rb
```diff
@@ -28,11 +29,11 @@ def start_element(name, attrs = [])
...
       when 'violation'
         @current_violation = PmdViolation.new(attrs, @branch_name)
       when 'error'
-        @current_filename = attrs['filename']
+        @current_filename = attrs['filename'].sub(/^#{@working_dir}/, '')
```
adangel: That's again only the hash key. `attrs['filename']` is not changed, so the full filename is later on still available in PmdError.

lib/pmdtester/pmd_report_detail.rb
```diff
@@ -25,6 +27,7 @@ def load(report_info_path)
         hash = JSON.parse(File.read(report_info_path))
         @execution_time = hash['execution_time']
         @timestamp = hash['timestamp']
+        @working_dir = hash['working_dir']
```
adangel: That would expose the working directory again.... which I though, you wanted to remove...

> The baseline is used for sharing, so we should remove this part of the information.

djydewang: This is used to record the working path when generating the pmd report. And we can remove the working dir from pmd report according to value of this 'working dir'

djydewang:
> I've created now the baseline locally and the full local working path is still contained in the pmd_report.xml files. Is this, what you wanted?

Yes, I agree with you that post-filtering the report should be a option of PMD itself.

> Note: I think, it won't work with PmdError, since the filename there is still with the local part (and used with the eql method).

Good catch! I'll fix it.

PR #26, opened 2018-07-16
Release pmdtester 1.0.0.beta1

djydewang:
- Adding a rake task named `hoe:spec` which can genenrate a gemspec based on a Hoe spec.

- Configuring travis CI's RubyGems Deployment


adangel:
That's actually merged...

PR #29, opened 2018-07-26
Add 'verify' rake task to verify code quality before committing changes

djydewang:
`rake verify` is equivalent to `rake clean test rubocop git:manifest bundler:gemfile`

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -7,8 +7,8 @@ require_relative './lib/pmdtester/parsers/options.rb'
 
 gem 'hoe'
 require 'hoe'
+
 Hoe.plugin :bundler
-Hoe.plugin :gemspec
 Hoe.plugin :git
 
 hoe = Hoe.spec 'pmdtester' do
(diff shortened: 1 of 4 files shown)


Rakefile
```diff
@@ -51,4 +51,6 @@ task 'hoe:spec' do
   File.open("#{hoe.name}.gemspec", "w") { |f| f.write hoe.spec.to_ruby}
 end
 
+desc 'verify code quality before committing changes'
+task 'verify' => ['clean', 'test', 'rubocop', 'bundler:gemfile', 'git:manifest']
```
adangel: `git:manifest` would generate the Manifest.txt file. Do we need to have this file then in the repository at all? Could we generate it just before we do `rake package` maybe?

Rakefile
```diff
@@ -7,8 +7,8 @@ require_relative './lib/pmdtester/parsers/options.rb'
 
 gem 'hoe'
 require 'hoe'
+
 Hoe.plugin :bundler
-Hoe.plugin :gemspec
```
adangel: Why don't we need hoe-gemspec anymore? I thought, we would require this for creating a ruby gem...

djydewang: We create a gemspec file by [hoe:spec](https://github.com/pmd/pmd-regression-tester/blob/bfe46f7d50f712830ece6ee8b63a7c080b1a6c76/Rakefile#L49) task now.

Rakefile
djydewang: Hm...Good idea. But there are some hoe tasks except `rake package` that may require manifest.txt, so I think it’s good to keep it all the time.

Rakefile
adangel: :+1:

PR #31, opened 2018-07-31
Fix color scheme for diff report, add default values for various options

djydewang:
- Fixing color scheme of the the diff report
- Adding default values to the option: `base_config`, `patch_config`, `list_of_project`

### DIFF
--- a/lib/pmdtester/parsers/options.rb
+++ b/lib/pmdtester/parsers/options.rb
@@ -14,6 +14,8 @@ module PmdTester
     LOCAL = 'local'
     ONLINE = 'online'
     SINGLE = 'single'
+    DEFAULT_CONFIG_PATH = ResourceLocator.locate('config/all-java.xml')
+    DEFAULT_LIST_PATH = ResourceLocator.locate('config/project-list.xml')
 
     attr_reader :local_git_repo
     attr_reader :base_branch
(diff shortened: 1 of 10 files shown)


lib/pmdtester/parsers/options.rb
```diff
@@ -111,20 +116,17 @@ def check_options
     def check_local_options
       check_option(LOCAL, 'base branch name', @base_branch)
       check_option(LOCAL, 'base branch config path', @base_config) unless @auto_config_flag
-      check_option(LOCAL, 'patch branch name', @patch_branch)
```
adangel: That explains why I have seen the log message twice on travis ... :+1:

lib/pmdtester/parsers/options.rb
```diff
@@ -111,20 +116,17 @@ def check_options
...
       check_option(LOCAL, 'patch branch config path', @patch_config) unless @auto_config_flag
       check_option(LOCAL, 'list of projects file path', @project_list)
     end
 
     def check_single_options
-      check_option(SINGLE, 'patch branch name', @patch_branch)
```
adangel: I think, we still need the patch_branch option, don't we? Ah... it's checked already in `check_common_options`...

PR #34, opened 2018-08-17
Clear old reports before generating new differences regression reports

djydewang:


lib/pmdtester/runner.rb
```diff
@@ -35,6 +36,13 @@ def run
       end
     end
 
+    def clean
+      clean_target = ['target/reports']
+      clean_target.each do |target|
```
adangel: Do we need a loop here? I'd prefer to keep it simple for now and only add the loop, if we need to clean more...

PR #39, opened 2018-08-21
Render stack traces properly

### LINKED ISSUE #37: Render stack traces properly
Hey @djydewang, great work on this project!

The stack traces eg [here](http://chunk.io/pmd/7cc8560a42ab43808202f57fe155eab4/diff/checkstyle/index.html) are not very readable because they lack line feeds, would it be possible to keep/reconstruct those?

djydewang:
Fixes #37

### DIFF
--- a/lib/pmdtester/builders/diff_report_builder.rb
+++ b/lib/pmdtester/builders/diff_report_builder.rb
@@ -214,7 +214,7 @@ module PmdTester
             doc.td pmd_error.msg
 
             # Details of error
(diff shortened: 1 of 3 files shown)


adangel:
Just one comment for your commit message: You have added "Fixes #37" to the PR description on github. That's great. But it would be better, to have this information also in the commit message. You can also use several lines for the commit message.
That way, a little bit more information is in the repository (although we still refer to github - so the number #37 does not help much, without looking it up on github...).

PR #44, opened 2018-09-05
Improve Danger messages & Increase the readability of the summary report

### LINKED ISSUE #38: Improve Danger messages
At the moment Danger messages are very basic, just saying that either no meaningful report is needed (ie: no Java changes), or to check out the report.

However, I believe we could do better...

1. State how many new violations were introduced
2. State how many violations were removed
3. State how many new errors were introduced
4. State how many error have been removed

So, we could have messages such as:

> This changeset introduces 3 new violations, and removes 1 error. [Full Report](xxx)

> This changeset removes 2 violations, adds 1 error and removes 2 errors. [Full Report](xxx)

> This changeset produces no differences. [Full Report](xxx)

That way, contributors have a better understanding of what may be wrong with their PR, and we know what to expect from the changeset. Otherwise, we end this PRs such as [this one](https://github.com/pmd/pmd/pull/1302#issuecomment-412639659), opening the report to see nothing changed.

djydewang:
Fixes #38


Gemfile
```diff
@@ -11,7 +11,7 @@ gem "hoe-bundler", "~>1.2", :group => [:development, :test]
 gem "hoe-git", "~>1.6", :group => [:development, :test]
 gem "minitest", "~>5.10.1", :group => [:development, :test]
 gem "mocha", "~>1.5.0", :group => [:development, :test]
-gem "rubocop", "~>0.56.0", :group => [:development, :test]
+gem "rubocop", "~>0.51.0", :group => [:development, :test]
```
adangel: I'll add a comment reminding us, that we use this specific version of rubocop for codacy (otherwise we might just update it in the future)

adangel: Actually we need to change it in the Rakefile and regenerate the Gemfile....

PR #45, opened 2018-09-10
Removes the exit statement in RuleSetBuilder Class

djydewang:
* Removes the exit statement in RuleSetBuilder Class
When pmdtester is used as a library, it will cause a lot of trouble for the caller if it contains an exit statement.
* Specify version of rdoc and rubygems


djydewang:
Good catch. The `-v` and `-h` options are mainly provided to the CLI, so users should avoid using these two options when calling pmdtester as a gem.

PR #52, opened 2018-10-09
Get the result of command 'java -version' from stderr rather than stdout

djydewang:
The` jdk version` in the summary table of [diff repor](http://chunk.io/pmd/bcbc33258efb46d1bc37906644c0afdb/diff/index.html) is empty. [Here](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4380614)'s why the result of the command `java -version` is going to stderr.

### DIFF
--- a/lib/pmdtester/cmd.rb
+++ b/lib/pmdtester/cmd.rb
@@ -6,6 +6,17 @@ module PmdTester
   # Containing the common method for executing shell command
   class Cmd
     extend PmdTester
+    def self.stdout_of(cmd)
+      stdout, _stderr, _status = execute(cmd)
+      stdout&.chomp!
+      stdout
+    end
+
+    def self.stderr_of(cmd)
+      _stdout, stderr, _status = execute(cmd)
+      stderr
+    end
+
     def self.execute(cmd)
       logger.debug "execute command '#{cmd}'"
 
@@ -20,7 +31,7 @@ module PmdTester
 
       stdout&.chomp!
 
-      stdout
+      [stdout, stderr, status]
     end
   end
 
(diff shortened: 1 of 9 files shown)


lib/pmdtester/cmd.rb
```diff
@@ -6,6 +6,17 @@ module PmdTester
...
+      stdout&.chomp!
+      stdout
+    end
+
+    def self.stderr_of(cmd)
+      _stdout, stderr, _status = execute(cmd)
+      stderr
+    end
+
     def self.execute(cmd)
```
adangel: How about changing the behavior of execute to return a combined string with stderr+stdout? That way, we wouldn't need to change the calls to execute everywhere....

You could keep your adjusted execute method and make it private and call it e.g. "internal_execute". And then provide the public method "execute" again with slightly adjusted behavior

lib/pmdtester/pmd_branch_detail.rb
```diff
@@ -26,8 +26,9 @@ def initialize(branch_name)
       branch_filename = PmdBranchDetail.branch_filename(branch_name)
       @base_branch_dir = "target/reports/#{branch_filename}" unless @branch_name.nil?
       @execution_time = 0
-      @jdk_version = Cmd.execute('java -version')
-      @language = Cmd.execute('echo $LANG')
+      # the result of command 'java -version' is going to stderr
+      @jdk_version = Cmd.stderr_of('java -version')
+      @language = Cmd.stdout_of('echo $LANG')
```
adangel: Instead of executing "echo $LANG", it's maybe easier to use just `ENV['LANG']`. Then we don't need to spawn a new process...

adangel:
Oh, and we should extend one of the integration tests to verify, that we correctly write the java version and LANG into the report. If we would have a test before, we would have seen already, that it didn't work :smile:

PR #53, opened 2018-10-10
Support comparing the two error stacktraces

### LINKED ISSUE #49: Support comparing two error stacktraces
In order to identify easier, whether a processing error is new or already existed before, we could highlight only the differences within a stacktrace.
So, it would be easier to spot, if only the line numbers changed, but the stacktrace is essentially the same.
Maybe, if only the line numbers have changed, we could skip this processing error for new/removed automatically.

Note: Partially address by #39

djydewang:
Fixes #49


lib/pmdtester/builders/diff_report_builder.rb
```diff
@@ -210,24 +213,38 @@ def build_errors_table_head(doc)
...
-            end
+        Array(0..(errors.size - 1)).each do |i|
+          build_errors_table_row(doc, i, errors)
+        end
+      end
+    end
 
-            # The error message
-            doc.td pmd_error.msg
+    def build_errors_table_row(doc, i, errors)
```
adangel: I would write this different - since the method does two things: either it outputs a diff if we have exactly to errors, one from base and one from patch. Otherwise it renders error by error.

I suggest to create two methods for that.
You can then call the methods in `build_errors_table_body` this way:

```
if ReportDiff.comparable?(errors)
    build_errors_table_row_diff(doc, errors[0], errors[1])
else
    errors.each do |pmd_error|
        build_error_table_row(doc, pmd_error)
    end
end
```

What do you think?

djydewang: Maybe we should do it in `build_errors_table_row` since there are duplicate logic in `build_errors_table_row_diff` and `build_error_table_row`.

PR #56, opened 2018-11-16
Pass the thread count down to the PmdReportBuilder

### LINKED ISSUE #46: Support multithreaded execution of PMD
When pmd-regression-tester executes PMD, it does not specify the `-threads` argument. This means, PMD is executed in single-thread-mode.

We maybe could speedup the analysis by using multi-threaded execution.

We can either specify the thread count here in pmd-regression-tester or change PMD, so that it determines the available CPUs itself and turns on multi-threading.

adangel:
Fixes #46

### DIFF
--- a/lib/pmdtester/runner.rb
+++ b/lib/pmdtester/runner.rb
@@ -35,10 +35,12 @@ module PmdTester
       return if rule_sets&.empty?
 
       PmdReportBuilder
-        .new(@options.base_config, @projects, @options.local_git_repo, @options.base_branch)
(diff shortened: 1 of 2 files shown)


djydewang:
@adangel Good catch! We should also pass the thread count in single mode https://github.com/pmd/pmd-regression-tester/pull/56/commits/39bce09ead36666e906c7897ce99a5d180e61b98#diff-7b9abf9b31952ff7f0aa2ee7d59af347R101

adangel:
Ah, yes - didn't think about single mode. Fixed :smile:

PR #59, opened 2018-11-17
Support configuration errors in the report

### LINKED ISSUE #57: Support configuration errors in the report
The report on pmd/pmd#1470 initially showed that violations had been removed, but that is because the rule had been removed from the ruleset as dysfunctional, because of the following changeset:

https://github.com/pmd/pmd/pull/1470/files#diff-4036d2a214e817321e2fb7694ea0aba2R241 

The report should have shown that new configuration errors were reported.

adangel:
Fixes #57

Handles configerror elements almost the same way like
other errors.


lib/pmdtester/builders/diff_report/configerrors.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def build_configerrors_table(doc, errors)
+    doc.table(class: 'bodyTable', border: '0') do
+      build_configerrors_table_head(doc)
+      build_configerrors_table_body(doc, errors)
+    end
+  end
+
+  def build_configerrors_table_head(doc)
```
djydewang: The structure of this method is similar to [build_errors_table_head](https://github.com/pmd/pmd-regression-tester/pull/59/files#diff-eb2dc6eb4abed74d6a76c2d3befa8a9fR27). Maybe we should define a method named `build_table_head(doc, head1, head2)` in `HtmlReportBuilder ` class. And the method should be:
```
def build_configerrors_table_head(doc)
    build_table_head(doc, 'Rule', 'Message')
end
```
`build_errors_table_head(doc)` should be:
```
def build_errors_table_head(doc)
    build_table_head(doc, 'Message', 'Details')
end
```

lib/pmdtester/builders/diff_report/configerrors.rb
```diff
@@ -0,0 +1,58 @@
...
+  def build_configerrors_table_row(doc, pmd_configerror)
+    doc.tr(class: pmd_configerror.branch == 'base' ? 'b' : 'a') do
+      build_configerrors_table_anchor(doc)
+
+      doc.td pmd_configerror.rulename
+      doc.td pmd_configerror.msg
+    end
+  end
+
+  def build_configerrors_table_anchor(doc)
```
djydewang: Same here. We could also define a method `build_table_anchor(doc, anchor_str, anchor_index)` in `HtmlReportBuilder` class since we do something similar in `build_configerrors_table_anchor`, `build_errors_table_anchor` and `build_violations_table_anchor`.

adangel:
@djydewang Good ideas, thanks for reviewing!

lib/pmdtester/builders/diff_report/configerrors.rb
adangel: Done.

lib/pmdtester/builders/diff_report/configerrors.rb
adangel: Done.

PR #64, opened 2020-09-28
[ci] Add github actions to execute ruby tests

oowekyala:
I think this might be nice. It uses the default template for ruby projects

### DIFF
--- a/ruby.yml
+++ b/ruby.yml
@@ -1,0 +1,29 @@
+# This workflow will download a prebuilt Ruby version, install dependencies and run tests with Rake
+# For more information see: https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby
+
+name: Ruby
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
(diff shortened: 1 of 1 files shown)


ruby.yml
```diff
@@ -0,0 +1,29 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby
+    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
+    # change this to (see https://github.com/ruby/setup-ruby#versioning):
+    # uses: ruby/setup-ruby@v1
+      uses: ruby/setup-ruby@ec106b438a1ff6ff109590de34ddc62c540232e0
```
adangel, line 23: I think, we should use `ruby/setup-ruby@v1` here....

adangel (approved):
I'll give it a try.

I think, the file needs to go to `.github/workflows/ruby.yml` - at least, the doc always uses this directory.

Note, that we already have a travis build, which also executes integration tests.

PR #65, opened 2020-09-29
Label PMD builds by sha and not by PMD version

### LINKED ISSUE #48: Support auxclasspath / typeresolution
We execute PMD just on the source code of standard projects now, but specifying the classpath for libraries used by the source code is also crucial for the type resolution.

This means, we probably need to build the project, we are analyzing, first.
And then figure out the full auxclasspath for the project.
We probably need then a small "build recipe" in addition to the source url for each project.

We could limit ourselves first for maven projects and execute `mvn compile` first. We still need to figure out the complete classpath though, which can be even more trickier for multi-module projects.

oowekyala:
This allows more opportunities to reuse previous PMD builds when running in local mode. Basically the target dir stores pmd distributions in directories labeled by the SHA of the commit from which they were built, instead of eg `pmd-bin-7.0.0-SNAPSHOT`. This means, if you have already built a distro on the correct commit, you can reuse it. Previously you could only reuse a build if you were on the correct branch and had a zip with the correct version in `pmd-dist/target`. So in local mode, where you generate a report for two branches, you need to build at least one branch every time you run the tool. And this also didn't necessarily mean that the zip had been built on the specific commit we wanted, just on a branch for the same pmd version.

Context: I'd like to fix #48 before we port java rules to PMD 7, and this change allows faster testing of this tool.


lib/pmdtester/builders/diff_report_builder.rb
```diff
@@ -89,9 +89,15 @@ def build_summary_row(doc, item, base, patch, *diff)
     end
 
     def build_filename_h3(doc, filename)
-      doc.h3 do
-        doc.a(href: @project.get_webview_url(filename)) do
-          doc.text @project.get_path_inside_project(filename)
+      if filename.nil?
```
adangel, line 92: I think, I've seen this problem locally as well now....

oowekyala: ill split this fix into a separate pr

oowekyala:
Notes about the draft status (to-dos for me):
* Integration tests have not been updated
* the build result in online mode should be reused, even if the build was not requested from danger (but by Travis)

oowekyala:
I don't think this is really needed anymore. Maybe we should pick it up later

PR #69, opened 2020-10-09
Detect single rules with auto-gen-config

adangel:
This allows a more fine-grained control of the dynamically generated ruleset when using `--auto-gen-config`.

The following is now possible:

* run regression tester on a pull request for pmd7 and compare it against pmd7-base
* run regression tester on a pull request for pmd7 and compare it against master. This now works, even though the baseline was created with different rules. In that case, we filter the baseline with the rules, that are actually used for the patch/PR build.

It's possible now in online mode, to overwrite (by specifying as cli options) the project list (which e.g. is different on master / pmd7) and the patch config ruleset (which is also different).

~It's not really well tested yet, unfortunately.~

I've tested it now locally and will soon test it with pmd/pmd#2800 :heavy_check_mark:


rsoesemann:
Dynamically generated rulesets? Cool when die that come?

adangel:
> Dynamically generated rulesets? Cool when die that come?

This is only for regression testing. We look at which files have been changed in the pull request and use a simple heuristic to find the rules, that are affected. Then we run PMD only with the changed rules against our two standard projects, which is faster, than running all rules.

Currently, this is only supported for java...

Do you know a open source apex project, that we could include here as well? See #2693 
Once we have a apex project, we can figure out, how to extend the logic for apex as well.

rsoesemann:
@adangel I recommend to use this huge and widely used apex repo https://github.com/apex-enterprise-patterns/fflib-apex-common

PR #74, opened 2020-11-04
Merge violations that have just changed messages

oowekyala:
We now attempt to find violations that have just changed messages, and instead of rendering them as 1 removed + 1 new, we output them as a single row with a different class. We also output a diff of the message

Also the "priority" column is removed from the html as it is useless

I don't know how to put a preview online, but here's a sample:
Capture d’écran de 2020-11-04 11-13-07
Where "java.lang" was inserted

The summary looks like this now:
Capture d’écran de 2020-11-04 11-18-19


adangel (approved):
Cool.

One thing (I'll add this): we call the regressiontester from Dangerfile and there we call "runner" and it returns the numbers (new, removed violations, etc.) that we use for the PR comment. We'll need to add the changed violations there as well.

PR #77, opened 2020-11-23
Better html reports

### LINKED ISSUE #73: Better HTML presentation for diff report
The table is very bulky and not helpful, if you want to look at a specific rule/file.

The idea is, to have a table, that can be
* filtered (by rules, by file names)
* sorted
* grouped (rows can be grouped, groups can be collapsed, expanded)

If there are many, many violations in the diff report, the file might also be very very big, potentially make your browser explode.
We could lazily fetch/load additional results. That would increase the initial load time.

Another useful feature would be, to fetch the code snippets and integrate them directly in the report. Currently you have a link to the github code, where the violation occurred, but that means, you need another browser tab. The code snippet could be shown by expanding a row. The overall size of the report must be considered: if we directly embed the code snippets, then the report is ever bigger. So, maybe just integrate it via iframe/fetch on demand.

One possible solution could be, to use https://datatables.net/

adangel:
Fixes #73


adangel:
@oowekyala I've create a PR from your branch. Unfortunately I don't have write access to push a fixed merge with master... Could you enable the flag "Allow edits and access to secrets by maintainers" please?
Next I'd like to review it and eventually merge it, so that all PRs on main pmd repo can benefit from this improvement.

oowekyala:
Thanks for opening this. I pushed a few more commits I had in local and a merge. Unfortunately I can't find an option to give you push rights on the branch. Maybe I need to open this myself

oowekyala:
Note that
* this is not thoroughly tested, I just made sure the current tests pass. But now the output is split into several files (an index.html and a project-data.js), and only index.html is tested
* this completely ignores config errors, they're not rendered or counted anymore, though I think, this can be understood if we remove them in pmd 7

PR #88, opened 2021-03-29
Fix html escaping in code snippets

oowekyala:
Creating elements using dom methods lets the browser escape the text content.

### DIFF
--- a/resources/js/code-snippets.js
+++ b/resources/js/code-snippets.js
@@ -1,7 +1,7 @@
 /**
  * Simple module to fetch code snippets via ajax.
  * Only supports github hosted code repos.
- * 
+ *
  * Usage:
  * var el = document.createElement('p');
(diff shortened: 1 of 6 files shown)


resources/js/code-snippets.js
```diff
@@ -11,53 +11,60 @@
...
+
+                container.appendChild(lineElt); // append to the container
             });
-            el.innerHTML = html;
         });
         oReq.open("GET", requestUrl);
         oReq.send();
 
-        el.innerHTML = "<tt>fetching...</tt>";
+        container.innerHTML = "<samp>fetching...</samp>";
```
oowekyala, line 67: (tt is deprecated)

adangel (approved):
Looks good.
This could be done as a bugfix release 1.1.2 (master is currently already on version 1.2.0-SNAPSHOT, but there are no commits yet).

PR #89, opened 2021-05-24
Make it possible to select a subpath of cloned directory

oowekyala:
The schema is updated to allow specifying a subpath within the cloned directory for pmd's `-d` option.

This will allow using the JDK sources as a regression project. Since pmd 7, it's reasonably fast to process eg the `java.base` module (2800 classes), while other modules are less interesting. If we want to select which modules we want to process, we need to use the `-d` option sensibly.

This also makes some optimisations to the git clone command (we don't fetch repo history anymore).

Btw since openjdk is now on Github, I think we could drop support for mercurial repos entirely. I assume that was meant for openjdk at the time.


oowekyala:
@adangel apparently the build script fails: https://github.com/pmd/pmd-regression-tester/pull/89/checks?check_run_id=2666316095#step:6:50 see here https://github.com/pmd/pmd-regression-tester/blob/master/.ci/build.sh#L19

I might just merge this and test with the push build, and release manually

adangel:
> I might just merge this and test with the push build, and release manually

I suspect, that it won't work either. We have to fix it....

For some reason, bundler is not found, but it has just been installed....

The last run on master branch was successful: https://github.com/pmd/pmd-regression-tester/runs/2517819162?check_suite_focus=true#step:6:48

They updated the docker image from https://github.com/actions/virtual-environments/blob/ubuntu20/20210504.1/images/linux/Ubuntu2004-README.md to https://github.com/actions/virtual-environments/blob/ubuntu20/20210517.1/images/linux/Ubuntu2004-README.md but I don't see any obvious change here....

adangel:
Maybe it's about this warning:

> WARNING:  You don't have /home/runner/.gem/ruby/2.7.0/bin in your PATH,
>  	  gem executables will not run.

adangel:
Maybe we never used the bundler we just installed, but the bundler, that was already there in the system, and now maybe it's gone?

Anyway #90 looks promising....

oowekyala:
> I suspect, that it won't work either. We have to fix it....

You're right of course ^^ I was thinking lazy

Apparently #90 fixes the problem. The build fails because of my code now

Thanks :)

adangel (commented):
I'll look into the problems. I think, the error "pathspec 'checkstyle-8.10' did not match any file(s) known to git" comes now, because we clone with "depth 1" only. And we have checkstyle in multiple project-list.xml files. And once without specifying a tag, so that we clone "master", and next with the tag "checkstyle-8.10". But if we already cloned the repo, we just try to switch branches, which fails now....

adangel:
I think this commit 5221212c3c28bb8c392895d20bd98cfdae6addcc solves the problem with "pathspec 'checkstyle-8.10' did not match any file(s) known to git". When using "--depth 1", the option "--single-branch" is implicitly enabled and needs to be explicitly disabled to clone all tags.

It might still fail, if we cached the cloned directory in github actions - so we might need to explicitly throw away the caches (by changing the cache key), so that pmd-regression-tester clones the projects from fresh...

adangel:

openjdk-11 is now analyzed: https://chunk.io/pmd/2867956b059144beaa2ae55e85c53079/diff1/openjdk-11/index.html
In this case, only one rule (AbstractClassWithoutAbstractMethod) is executed, but the report is generated.

Note: In the baselines, this project doesn't exist yet, so all violations are considered new for now.

oowekyala:
Great, thanks for picking this up :)

PR #91, opened 2021-06-17
Filter baseline based on patch config

adangel:
This addresses the problem described in pmd/pmd#3336

> When executing regression tester against master, we
> don't actually run PMD again, but we use the result from
> the previous run. There only the modified rules have been
> executed, so we should use the same rules for comparison
> against master.
> This should result in smaller reports, since the comparison
> on master reported all the not executed rules as missing
> violations.

The new cli option `--filter-with-patch-config` looks at the patch config and creates a filterset with only the rules enabled in the patch config. Then this filter is used when generating the diff to the baseline, which avoids reporting violations as removed, when the corresponding rules were actually not active.

That should help us in interpreting PMD 7 rule updates more easily.

Next steps:

* create a new PR against pmd/7.0.x with the following changes :heavy_check_mark: pmd/pmd#3342
  * Gemfile -> reference this branch
  * Dangerfile -> see pmd/pmd#3336 for the changes, instead of `--auto-gen-config` use `--filter-with-patch-config`
  * Change a single java rule
* Verify regression tester report :heavy_check_mark:


adangel:

Instead of removing 304620 violations, now only 147 violations are removed when compared to master. The report is only about the changed rule now.

PR #92, opened 2021-06-17
Include full pmd report

### LINKED ISSUE #71: Include full PMD report
Currently we only show the differences. This assumes, that the baseline is the ground truth. But that might not always be correct.

It would be helpful, if we additionally upload the full report, so one can skim through to verify, that we report correct violations.

Also having a look at any processing error would be beneficial.

Maybe we could even share the logs from the pmd run (warnings, etc.). Currently, the logs are only provided in the builds logs, if PMD execution failed.

adangel:
* Add a header to the generated pages (currently the reports just begin...) and a timestamp on the main index page (so that it's directly visible, whether the report is old or new)
* Fixes #71 

~That's a minimal solution - we should now be able to download the full pmd_report.xml file (for both base and patch branches), so that we can have a look at the original report.~

~It addresses #71 only partially, because we should also render the pmd_report.xml files into a nice html page, so we can easily skim through the results in the browser. Maybe we can just run a XSLT transformation on these files additionally....~

The Full PMD Report is now render as a html page as well, using the same datatable view. Additionally, the original XML report can be downloaded as well.


adangel:

Full checkstyle report: https://chunk.io/pmd/2867956b059144beaa2ae55e85c53079/diff2/checkstyle/base_pmd_report.html

Note, that the rule set filter is also applied here. Actually this happened accidentally- I didn't think about it. But for now, I think, that's ok.

The original pmd xml file (https://chunk.io/pmd/2867956b059144beaa2ae55e85c53079/diff2/checkstyle/base_pmd_report.xml) contains all violations from the baseline - so that's the really complete violations list.

PR #102, opened 2021-12-10
Use long cli options for PMD >= 6.41.0

adangel:
This is to avoid deprecation warnings.


lib/pmdtester/builders/pmd_report_builder.rb
```diff
@@ -194,5 +195,16 @@ def saved_distro_path(build_sha)
...
       !Cmd.execute('git status --porcelain').empty?
     end
+
+    def should_use_long_cli_options
+      logger.debug "PMD Version: #{@pmd_version}"
+      m = /(\d+)\.(\d+)\.(\d+).*/.match(@pmd_version)
+      major = m[1]
+      minor = m[2]
+      patch = m[3]
+      result = major.to_i >= 6 && minor.to_i >= 41
```
jsotuyod, line 205: this will fail with PMD 7.0.0 as `inor.to_i >= 41` will be false

adangel: you're right :+1:

PR #103, opened 2022-03-18
Support other languages besides java

adangel:
This changes the regression tester to support other languages than java.
The main changes summarized:

* When determining the rules from the changed files, we filter the changed files by language (to ignore changes e.g. in pmd-plsql which don't affect regression testing). The list of languages is determined from the patch config (ruleset).
* When we determine the rules/rulesets that are to be run (e.g. for automatically generating a ruleset), the rule references are prefixed with the language. E.g. instead of `bestpractices.xml/AbstractClassWithoutAbstractMethod` it will now be `java/bestpractices.xml/AbstractClassWithoutAbstractMethod`.
* The same is used for filtering the results (filter_set) - there will always be the language now
* To determine the language from a given violation, the externalInfoUrl is used. It seems, we don't output the language in the XML...

This should allow us to enable regression testing for Apex as well (see #2693).

The baseline with 3 apex and 3 java projects looks like this:
grafik


lib/pmdtester/builders/rule_set_builder.rb
```diff
@@ -158,5 +118,82 @@ def write_dynamic_file(rule_refs)
...
+        break
+      end
+      [regression_test_required, categories, rules]
+    end
+
+    def check_single_filename(filename, categories, rules)
+      logger.debug "Checking #{filename}"
+
+      # matches Java-based rule implementations
+      match_data = %r{.+/src/main/java/.+/lang/([^/]+)/rule/([^/]+)/([^/]+)Rule.java}.match(filename)
```
adangel, line 147: This is the place, where we try to map from the changed files in the PR to the changed rules. But this regex works even if the projects are a bit restructured, as proposed in pmd/pmd#3846 . As long as the structure inside a language stays...

test/test_rule_set_builder.rb
```diff
@@ -47,11 +48,25 @@ def test_build_design_codestyle_config
...
-    mock_build(diff_filenames, nil, 'my-patch-config.xml')
+    mock_build(diff_filenames, nil, "#{PATH_TO_TEST_RESOURCES}/patch-ruleset.xml")
+
+    assert(!File.exist?(RuleSetBuilder::PATH_TO_DYNAMIC_CONFIG),
+           "File #{RuleSetBuilder::PATH_TO_DYNAMIC_CONFIG} must not exist")
+  end
+
+  def test_build_all_rulesets_config_salesforce
+    diff_filenames = <<~DOC
+      pmd-salesforce/pmd-apex/src/main/java/net/sourceforge/pmd/lang/apex/rule/bestpractices/AvoidGlobalModifierRule.java
```
adangel, line 64: That's an example how the project structure could look like

PR #113, opened 2022-05-09
Fix missing fetch

### LINKED ISSUE #106: git clone/checkout fails when using commit sha1 as tag
Currently we need to fork [Schedul-o-matic-9000](https://github.com/pmd/Schedul-o-matic-9000) and [fflib-apex-common](https://github.com/pmd/fflib-apex-common) in order to create a tag, that we can use in [project-list.xml](https://github.com/pmd/pmd/blob/master/.ci/files/project-list.xml):

```xml
  <project>
    <name>Schedul-o-matic-9000</name>
    <type>git</type>
    <connection>https://github.com/pmd/Schedul-o-matic-9000</connection>
    <!-- 
    <connection>https://github.com/SalesforceLabs/Schedul-o-matic-9000</connection>
    Note: using forked repo from pmd, as original repo doesn't have any useful tags/branches
    -->
    <tag>pmd-regression-test</tag>
  </project>
```

It would be nice to allow to specify directly a sha1 for the commit, e.g.

```xml
  <project>
    <name>Schedul-o-matic-9000</name>
    <type>git</type>
    <connection>https://github.com/SalesforceLabs/Schedul-o-matic-9000</connection>
    <tag>6b1229ba43b38931fbbab5924bc9b9611d19a786</tag>
  </project>
```

Using the `tag`-tag would avoid to change the schema for this file.

oowekyala:
Fix #106 

Ref problems in pmd/pmd#3960

### DIFF
--- a/lib/pmdtester/builders/project_builder.rb
+++ b/lib/pmdtester/builders/project_builder.rb
@@ -30,9 +30,17 @@ module PmdTester
         end
 
         Dir.chdir(path) do
-          # this should work with tags, branch names and (full-length) hashes
+          # this works with tags, branch names and (full-length) hashes
+          # first move to a different (temporary) branch, in case we are already on the branch that we want to update
+          Cmd.execute_successfully('git checkout -b fetched/temp')
+          # fetches any new commits. Could also be a tag.
           Cmd.execute_successfully("git fetch --depth 1 origin #{project.tag}")
-          Cmd.execute_successfully("git checkout #{project.tag}; git reset --hard #{project.tag}")
+          # update the branch to work on based on the new fetch. Creates a new branch if it doesn't exist already
+          Cmd.execute_successfully("git branch --force fetched/#{project.tag} FETCH_HEAD")
+          # checkout the updated branch
+          Cmd.execute_successfully("git checkout fetched/#{project.tag}")
+          # remove the temporary branch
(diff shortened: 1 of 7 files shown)


lib/pmdtester/builders/project_builder.rb
```diff
@@ -19,22 +19,22 @@ def clone_projects
...
-
+          # Don't download whole history. This just fetches HEAD, the correct ref is fetched below.
+          clone_cmd = "git clone --single-branch --depth 1 #{project.connection} #{path}"
           Cmd.execute_successfully(clone_cmd)
         end
 
         Dir.chdir(path) do
-          execute_reset_cmd(project.type, project.tag)
+          # this should work with tags, branch names and (full-length) hashes
+          Cmd.execute_successfully("git fetch --depth 1 origin #{tag}")
```
oowekyala, line 36: note: i haven't actually checked whether this works for tags. the proper syntax is maybe `tags/tagname`

oowekyala: See https://stackoverflow.com/questions/45338495/fetch-a-single-tag-from-remote-repository

adangel: I'll look at this - I like the idea to use a fetch here. This will make it easier to reuse the clones, even if upstream changes.

PR #117, opened 2022-11-25
Use relative paths without leading slash

### LINKED ISSUE #116: Relative paths in pmd report result in missing code snippets
PMD6 reports absolute paths, PMD7 seems to report only relative paths (relative to the current working directory).

* pmd6: `/home/runner/work/pmd/target/repositories/openjdk-11/src/java.base/aix/classes/sun/nio/ch/AixAsynchronousChannelProvider.java`
* pmd7: `target/repositories/openjdk-11/src/java.base/aix/classes/sun/nio/ch/AixAsynchronousChannelProvider.java`

The regression tester should be able to deal with both situations. It will relativize the paths on its own (relative to the project root, that is `/home/runner/work/pmd/target/repositories/openjdk-11`).

The problem is, that we generate wrong links to the source code, e.g. https://github.com/openjdk/jdk/tree/jdk-11+28/target/repositories/openjdk-11/src/java.base/share/classes/java/util/AbstractMap.java#L609 instead of https://github.com/openjdk/jdk/tree/jdk-11+28/src/java.base/share/classes/java/util/AbstractMap.java#L609

adangel:
Fixes #116

### DIFF
--- a/lib/pmdtester/parsers/pmd_report_document.rb
+++ b/lib/pmdtester/parsers/pmd_report_document.rb
@@ -41,11 +41,11 @@ module PmdTester
     end
 
     def characters(string)
-      @cur_text << remove_work_dir!(string)
+      @cur_text << string
     end
 
     def cdata_block(string)
-      @cur_text << remove_work_dir!(string)
+      @cur_text << string
     end
 
     def end_element(name)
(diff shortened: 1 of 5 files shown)


adangel:
The fix has been successfully tested with https://github.com/pmd/pmd/pull/4221 - now the comparison between master and pmd/7.0.x works again and the code snippets are fetched again.

PR #122, opened 2023-11-09
Escape violation messages

### LINKED ISSUE #121: Violation messages should be escaped for html
When rule messages contain generic types (e.g. `List<X>`), then X is not displayed, because the angle brackets are not displayed in interpreted by the browser as tags:

image

adangel:
Fixes #121

### DIFF
--- a/lib/pmdtester/builders/project_hasher.rb
+++ b/lib/pmdtester/builders/project_hasher.rb
@@ -117,7 +117,7 @@ module PmdTester
         'l' => violation.line,
         'f' => file_ref,
         'r' => violation.rule_name,
-        'm' => create_violation_message(violation, is_diff)
+        'm' => create_violation_message(violation, is_diff && violation.changed?)
       }
       h['ol'] = violation.old_line if is_diff && violation.changed? && violation.line != violation.old_line
       h
@@ -132,8 +132,6 @@ module PmdTester
     end
 
     def escape_html(string)
-      return '' if string.nil?
-
       CGI.escapeHTML(string)
     end
   end
--- a/lib/pmdtester/builders/project_hasher.rb
+++ b/lib/pmdtester/builders/project_hasher.rb
@@ -117,15 +117,24 @@ module PmdTester
         'l' => violation.line,
         'f' => file_ref,
         'r' => violation.rule_name,
-        'm' => is_diff && violation.changed? ? diff_fragments(violation) : violation.message
+        'm' => create_violation_message(violation, is_diff)
       }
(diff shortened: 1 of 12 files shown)


adangel:
This is fixed now:

image

lib/pmdtester/builders/project_hasher.rb
```diff
@@ -117,15 +117,24 @@ def make_violation_hash(file_ref, violation, is_diff = TRUE)
...
+        'm' => create_violation_message(violation, is_diff)
       }
       h['ol'] = violation.old_line if is_diff && violation.changed? && violation.line != violation.old_line
       h
     end
 
-    def diff_fragments(violation)
-      diff = Differ.diff_by_word(violation.message, violation.old_message)
+    def create_violation_message(violation, is_diff)
+      return escape_html(violation.message) unless is_diff
```
jsotuyod, line 127: the old code checked not only against `is_diff`, but also for `violation.changed?`… is this change intentional?

lib/pmdtester/builders/project_hasher.rb
```diff
@@ -117,15 +117,24 @@ def make_violation_hash(file_ref, violation, is_diff = TRUE)
...
+
+      diff = Differ.diff_by_word(escape_html(violation.message),
+                                 escape_html(violation.old_message))
```
jsotuyod, line 130: thinking out loud here, but is it ok to diff the html? or should we escape the raw diff? I can't think of a reason why it wouldn't be equivalent, but I'm not sure if it would hold for all scenarios…

lib/pmdtester/builders/project_hasher.rb
adangel: Good catch! I think we need to check for both (although I didn't see a difference in behavior yet). Only when violation.changed? is true, we would have a violation.old_message available. Maybe that explains, why I needed line 135 to deal with nil...

lib/pmdtester/builders/project_hasher.rb
adangel: The result of Differ.diff_by_word is itself already html - so we need to escape it before, e.g. it produces something like this:

```html
The method 'triggerHandler(Type, Boolean, Boolean, Boolean, Boolean, Boolean, Boolean, List&lt;SObject&gt;, Map&lt;Id<del class="differ">,SObject&gt;)' has a cognitive complexity of 25</del>, <ins class="differ">SObject&gt;)' has a cognitive complexity of 25, </ins>current threshold is 15
```

It is adding `<del>` and `<ins>` tags...

PR #123, opened 2024-02-15
Support ruby 3.3.0

adangel:
And also update rubocop

See also pmd/pmd#4776


adangel:
This doesn't mean, that pmd-regression-tester requires ruby 3.3.0. It just means, we don't test anymore, that it runs with ruby 2.7...

.github/workflows/build.yml
```diff
@@ -31,16 +31,17 @@ jobs:
...
-        ruby-version: 2.7
+        ruby-version: 3.3
     - name: Setup Environment
       shell: bash
       run: |
         echo "LANG=en_US.UTF-8" >> $GITHUB_ENV
         echo "MAVEN_OPTS=-Dmaven.wagon.httpconnectionManager.ttlSeconds=180 -Dmaven.wagon.http.retryHandler.count=3" >> $GITHUB_ENV
-        echo "PMD_CI_SCRIPTS_URL=https://raw.githubusercontent.com/pmd/build-tools/17/scripts" >> $GITHUB_ENV
+        #echo "PMD_CI_SCRIPTS_URL=https://raw.githubusercontent.com/pmd/build-tools/master/scripts" >> $GITHUB_ENV
+        echo "PMD_CI_SCRIPTS_URL=https://raw.githubusercontent.com/adangel/build-tools/ruby3/scripts" >> $GITHUB_ENV
```
jsotuyod, line 44: if we merge https://github.com/pmd/build-tools/pull/29 before I don't think we need this…

PR #126, opened 2024-09-05
Fix integration tests

adangel:
Add a maven-settings.xml for building (old) PMD versions with another maven repository, since dokka-maven-plugins uses a dependency (kotlinx-html-jvm) which was once available via jcenter.bintray.com, but bintray.com no longer exists today.

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -56,6 +56,7 @@ resources/js/datatables.min.js
 resources/js/jquery-3.6.4.slim.min.js
 resources/js/popper-2.11.7.min.js
 resources/js/project-report.js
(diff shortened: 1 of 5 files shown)


jsotuyod:
I'm a little confused here…

If I understand correctly, the issue lies with the `dokka-maven-plugin` declared by `pmd-lang-test` referencing a dependency (`kotlinx-html-jvm`) that is not available in maven central, but on the now defunct jcenter… now, if that's the case, shouldn't this also fail when building `pmd-lang-test` directly from the pmd repo? shouldn't this be fixed there rather than here?

PR #141, opened 2025-09-08
Fix #124: If an abstract rule is changed, run all rules

### LINKED ISSUE #124: With `--auto-gen-config` invalid rule references might be generated
See https://github.com/pmd/pmd/pull/4971 for an example:

The changeset contains a change to file `pmd-java/src/main/java/net/sourceforge/pmd/lang/java/rule/codestyle/AbstractNamingConventionRule.java`, which is not a rule by itself but a base class.

The regression tester however generates a ruleset like that:

```
[ERROR] Error at target/reports/HEAD/checkstyle/config.xml:8:5
  6|     <rule ref="category/apex/design.xml/CognitiveComplexity"/>
  7|     <rule ref="category/apex/design.xml/CyclomaticComplexity"/>
  8|     <rule ref="category/java/codestyle.xml/AbstractNamingConvention"/>
         ^^^^^ Unable to find referenced rule AbstractNamingConvention; perhaps the rule name is misspelled?

  9|     <rule ref="category/java/codestyle.xml/FieldNamingConventions"/>
 10|     <rule ref="category/java/codestyle.xml/IdenticalCatchBranches"/>
[ERROR] Cannot load ruleset target/reports/HEAD/checkstyle/config.xml: An XML validation error occurred
```

But a rule "AbstractNamingConvention" doesn't exist...

The result is, that PMD is not executed at all and the diff report shows, that all violations are missing.

UncleOwen:
Since I ran into this twice in short succession, I figured I might just as well fix this.

Applies a simple heuristic: If the name of a rule starts with 'Abstract', it's an abstract class. This could change any number of rules, so let's bail and just compare all rules.

My knowledge of ruby is very basic. I basically copied one of the other "compare all rules" test cases.

Fixes #124

### DIFF
--- a/lib/pmdtester/builders/rule_set_builder.rb
+++ b/lib/pmdtester/builders/rule_set_builder.rb
@@ -146,7 +146,7 @@ module PmdTester
 
       # matches Java-based rule implementations
       match_data = %r{.+/src/main/java/.+/lang/([^/]+)/rule/([^/]+)/([^/]+)Rule.java}.match(filename)
-      unless match_data.nil?
+      unless match_data.nil? || match_data[3].start_with?('Abstract')
         logger.debug "Matches: #{match_data.inspect}"
         rules.add("#{match_data[1]}/#{match_data[2]}.xml/#{match_data[3]}")
         return true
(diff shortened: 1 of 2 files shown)


adangel (approved):

UncleOwen:
@zbynek I know. See the resolved discussion with adangel above.

